PR #1, opened 2019-10-08
[ADDED] added files for a near-blank React-Native app 

MPucheta:
- Main problem was: Expo vs React-Native init. Expo is generally easier to setup and to prototype with, but is limited in functionality regarding native modules and 3rd API's, so ejecting would be required. Given we can't predict how this will evolve, and the many blog posts telling that Expo ejecting immediately led to downtime in development (dependencies breaking, problems on building) 'react-native init' seems safer. This impact folder structure and files tho, I do not believe it's too important, for now...


abelosorio (approved):
Looks good to me.

nahudalla (approved):

PR #2, opened 2019-10-09
Mp/2.create the signup screen

MPucheta:
- Added signup screen. It connects to firebase using credentials on a .env, so that we do not put keys in the code. I also implemented persistant login, although It seems to be working in the emulator, I did not try it on my real phone. 

- Also added some simple screens to check if navigation was working, and also to prepare the code when I add more screens.

- Added a centralized navigation file, alternative was to clutter code with useContext hooks and such (which was breaking the component tree, as the render tests did not pass)

- I envision that some duplication will be present in login screen, I do not consider important to refactor it in this iteration.

image


src/screens/SignUp/index.js
```diff
@@ -0,0 +1,102 @@
...
+import * as firebase from 'firebase'
+
+const SignUp = () => {
+  const [authenticating, setAuthenticating] = useState(false)
+  const [error, setError] = useState(null)
+
+  const handleSubmit = (values) => {
+    setAuthenticating(true)
+    firebase.auth().createUserWithEmailAndPassword(values.email, values.password)
+      .then(user => {
```
abelosorio: Can we use `async/wait` instead of then/catch?

abelosorio (changes requested):
Looks great! Just a single minor comment about the usage of async/wait.

src/App.js
```diff
@@ -0,0 +1,33 @@
...
+import * as firebase from 'firebase'
+import firebaseConfig from './configs/firebase'
+
+const initializeFirebase = () => {
+  if (!firebase.apps.length) {
+    firebase.initializeApp(firebaseConfig)
+  }
+}
+
+initializeFirebase()
```
nahudalla: Do you need to call this function again somewhere else? If you don't, then you can write the `if` statement at the top level, without a function. It'll be executed only once, the first time this module is imported by another one (or run as an entry point).

src/screens/Home/index.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { Text, View, Button } from 'react-native'
+import NavigationService from '../../navigationService'
+import * as firebase from 'firebase'
+
+const Home = () => {
+  const LogOut = async () => {
+    await firebase.auth().signOut()
+    NavigationService.navigate('SignUp')
+  }
```
nahudalla: If this function is not accessing anything form its closure, move its definition to the top level. That way it'll be created only once and not every time the Home component renders.

src/screens/LoadingScreen/index.js
```diff
@@ -0,0 +1,24 @@
...
+const LoadingScreen = () => {
+  const checkIfLoggedIn = () => {
+    firebase.auth().onAuthStateChanged(user => {
+      if (user) {
+        NavigationService.navigate('Home')
+      } else {
+        NavigationService.navigate('SignUp')
+      }
+    })
+  }
```
nahudalla: Same here about defining the function at the top level.

src/screens/SignUp/index.js
nahudalla: Yes, it's possible

src/screens/SignUp/index.js
```diff
@@ -0,0 +1,102 @@
...
+        setError(null)
+        setAuthenticating(false)
+        NavigationService.navigate('Home')
+      })
+      .catch(error => {
+        setError(error)
+        setAuthenticating(false)
```
nahudalla: When you need to do something in both cases (`then/catch`), use `finally`. This also applies to `try/catch` blocks:
```javascript
promise.then(() => {
  doSomethingOnlyWhenEverythingIsOK()
}).catch((err) => {
  doSomethingWhenAnErrorIsThrown(err)
}).finally(() => {
  alwaysDoThisNoMatterIfAnErrorWasThrownOrNot()
})

// with try/catch
try {
  await promise
  doSomethingOnlyWhenEverythingIsOK()
} catch (err) {
  doSomethingWhenAnErrorIsThrown(err)
} finally {
  alwaysDoThisNoMatterIfAnErrorWasThrownOrNot()
}
```

nahudalla (changes requested):

src/screens/SignUp/index.js
```diff
@@ -0,0 +1,118 @@
...
+  const [error, setError] = useState(null)
+
+  const handleSubmit = async (values) => {
+    setAuthenticating(true)
+    try {
+      await firebase.auth().createUserWithEmailAndPassword(values.email, values.password)
+      NavigationService.navigate('Home')
+    } catch (error) {
+      setError(error)
+      setAuthenticating(false)
```
nahudalla, line 21: My previous comment was removed by the latest commit. This statement should be in a `finally` block.

MPucheta: My bad! I was pushing -f as Abel requested a rebase. There is some problem though depending on which React Native navigator is being used. If using SwitchNavigator, then setting state in the finally block induces a warning as the component will be unmounted. If using StackNavigator, no warning appears (I believe the component is still mounted, but the docs are not clear in this aspect)

nahudalla: That's weird... Both `catch` and `finally` should execute one right after the other... Even before the component is unmounted... I don't know though, I'm not an expert in React Native. It may be possible for the component to be unmounted...

nahudalla (approved):
A rebase is still needed

src/screens/SignUp/index.js
MPucheta: Yes indeed. Might be something on the navigation. It could be that if you are navigating somewhere in the component tree, then all others (that are on different "branches") gets signaled for "WillBeUnmouted" thus giving the warning.

src/App.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { createAppContainer, createSwitchNavigator } from 'react-navigation'
+import NavigationService from './navigationService'
+import SignUp from './screens/SignUp'
+import Home from './screens/Home'
+import LoadingScreen from './screens/LoadingScreen'
+import * as firebase from 'firebase'
+import firebaseConfig from './configs/firebase'
+
+if (!firebase.apps.length) {
```
abelosorio, line 10: I think anyone can understand this condition at a glance. However, if you try to read it "in English", it would be: "if not length" or "if the length if false", even though we understand this is *interpreted* as "if the length evaluates to false*.

It's not a big deal, but I think a more readable approach could be `if (firebase.apps.length === 0)` or `if (!firebase.apps)`.

src/screens/Home/index.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import { Text, View, Button } from 'react-native'
+import NavigationService from '../../navigationService'
+import * as firebase from 'firebase'
+
+const LogOut = async () => {
+  await firebase.auth().signOut()
```
abelosorio, line 7: Watch out for asynchrony calls outside of try/catch blocks. They all should be wrapped in a try/catch block and, if you don't want to react to that error different than the original error, you can just re-throw it.
```javascript
try {
  await firebase.auth().signOut()
} catch (err) {
  throw err
}
```

abelosorio (approved):
Great job! A few non-blocker comments ;)

PR #4, opened 2019-10-16
MP/4.Add username information to SignUp and Login

MPucheta:
- Added a bottom tab navigator for future screens, with this change I may be able to do the screens independently (To continue working with them as separate features instead of coding on top of the previous branch, and also reduce the number of different files I change per pull). **As for now, this pull must be merged after MP/3**

- Added Username to SignUp and Login. Now we can login with the username too (and as such, at the moment, the usernames must also be unique).

- I did quite a few tests with the database as I will probably use it later and I wanted to do an abstraction, regardless of this, the code retains some verbosity. A refactoring may be useful at the end of the iteration.

**Changes after refactoring**

- **Screens are now separated into more granular components, each with their own styles. Some things could be shared, such as Background and AuthenticatingIndicator, others such as Headers and AccountLink/NoAccountLink can be abstracted but in doing so they result confusing to read.**

- **Changed database to cloud firestore (noSQL based on documents)**

- **I had an urge of abstracting firebase.auth(), firebase.firestore() and initialization into a single file for modularity. It seemed a little overkill as the firebase API is very clear to use and read and another layer could have the adversary effect, what do you think?**

- I attach some pics

image

image

### DIFF
--- a/src/databaseService/index.js
+++ b/src/databaseService/index.js
@@ -1,0 +1,57 @@
+import 'react-native'
+import * as firebase from 'firebase'
+
+const set = async (docPath, data) => {
+  try {
+    await firebase.firestore().doc(docPath).set(data)
+  } catch (error) {
+    console.log(error.message)
+  }
+}
+
+const get = async (docPath) => {
+  try {
+    const databaseSnapshot = await firebase.firestore().doc(docPath).get()
+
+    return databaseSnapshot.data()
+  } catch (error) {
+    console.log(error.message)
+  }
+}
+
+const update = async (docPath, data) => {
+  try {
+    await firebase.firestore().doc(docPath).update(data)
+  } catch (error) {
+    console.log(error.message)
+  }
+}
+
+const queryEmailFromUsername = async (username, collectionPath = 'users') => {
+  try {
+    const querySnapshot = await firebase.firestore().collection(collectionPath).where('username', '==', username).get()
+    if (hasDocuments(querySnapshot)) {
+      const json = getFirstDocumentDataFromQuerySnapshot(querySnapshot)
+      return json && json.email
+    }
+  } catch (error) {
+    console.log(error.message)
+  }
+}
+
+const getFirstDocumentDataFromQuerySnapshot = (querySnapshot) => querySnapshot.docs[0].data()
+const hasDocuments = (querySnapshot) => querySnapshot.size > 0
+
+const usernameAlreadyInUse = async (username, collectionPath = 'users') => {
+  const email = await queryEmailFromUsername(username, collectionPath)
+
+  return email !== undefined
+}
+
+export default {
+  set,
+  get,
+  update,
+  usernameAlreadyInUse,
+  queryEmailFromUsername
+}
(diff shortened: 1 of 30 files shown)


src/databaseService/index.js
```diff
@@ -0,0 +1,59 @@
...
+const set = async (refPath, data) => {
+  try {
+    await firebase.database().ref(refPath).set(data)
+  } catch (error) {
+    console.log(error.message)
+  }
+}
+
+const get = async (refPath, mode = 'value') => {
+  let values = null
```
abelosorio: I think you don't need this variable. You can just return `databaseSnapshot.val()`.

src/databaseService/index.js
```diff
@@ -0,0 +1,59 @@
...
+  return json[Object.keys(json)[0]].email
+}
+
+const queryEmailFromUsername = async (username, refPath = 'users/') => {
+  const ref = firebase.database().ref(refPath)
+  let email = null
+  try {
+    const databaseSnapshot = await ref.orderByChild('username').equalTo(username).once('value')
+    const json = databaseSnapshot.val()
+    email = json && getEmailFromJson(json)
```
abelosorio: Same here, you can do instead: `return json && getEmailFromJson(json)`

src/databaseService/index.js
```diff
@@ -0,0 +1,59 @@
...
+  } catch (error) {
+    console.log(error.message)
+  }
+  return email
+}
+
+const usernameAlreadyInUse = async (username, refPath = 'users/') => {
+  const email = await queryEmailFromUsername(username, refPath)
+  return email !== null
```
abelosorio: Let's leave a blank line before return statements (when they are among other statements within the same block).

src/screens/authentication/Login/index.js
```diff
@@ -0,0 +1,110 @@
+import React, { useState } from 'react'
```
abelosorio, line 1: The Login and Signup components are quite long. I'd suggest to split them out into smaller and reusable files. For example:
```javascript
    <LoginBackground>
      <LoginHeader />
      <LoginForm
        initialValues={initialValues}
        errors={errors}
        authenticating={authenticating}
        onSubmit={handleSubmit}
      />
      <NoAccountLink />
    </LoginBackground>
```

src/screens/authentication/utils/index.js
```diff
@@ -0,0 +1 @@
+export const isNotEmail = (text) => !text.includes('@')
```
abelosorio: A very small detail, but I had to come here to see what `isNotEmail` was. I'd suggest renaming to `isValidEmail`. @nahudalla what do you say?

abelosorio (commented):
Awesome!! It's coming along! A few small comments, tho.

src/screens/authentication/utils/index.js
nahudalla: Yes, I agree. Also, this is not a correct way of checking if it's not a valid email address... but I think it's OK for now

nahudalla (approved):

src/screens/authentication/utils/index.js
MPucheta: You are correct, I considered if putting a RegExp was necessary as a badly formatted email would not be logged by firebase and only needed the distinction between username and email. 

I will rethink if what I intended is a "**isUsername**" functionality instead.

nahudalla (approved):

abelosorio (approved):

PR #5, opened 2019-10-22
[MP/5] Implemented Diagnose Request Screen

MPucheta:
- A diagnose reference is added to the user's diagnose array.

- General project structure was improved. I also added an authenticationService for better separation of concerns, although I did not modify SignUp and Login, It's not this pull intent and it would be better if done in a separate one.

- Added timestamps. Only a portion of methods has the capability to add timestamps, they are the "ORM layer" and should be used instead of the firebase API ones.

- Added an error hierarchy, allowing for polymorphism to be achieved (we could implement a Visitor Pattern with an error handler too should we need it).

image

image

**Database**

image

**Database**

image

**Bucket**

image


src/authenticationService/index.js
```diff
@@ -0,0 +1,13 @@
+import * as firebase from 'firebase'
+
+const getCurrentUserUID = () => {
+  try {
+    return firebase.auth().currentUser.uid
+  } catch (error) {
+    throw new Error('Authentication problem: Not logged in.')
```
abelosorio: Instead of determining the error type by the message of it, we should have different types of error: `AuthenticationError`, `NetworkError`, etc, so that we can react to them properly.

src/blobService/index.js
```diff
@@ -0,0 +1,12 @@
+const buildBlobFromURI = async (uri) => {
+  try {
+    const fetchURI = await fetch(uri)
+    return fetchURI.blob()
+  } catch (error) {
+    throw new Error('Blob problem: could not build Blob from URI.')
```
abelosorio: Same here.

src/storageService/index.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    const blob = await BlobService.buildBlobFromURI(imageURI)
+    const uuid = uuidv1()
+    const imageRef = firebase.storage().ref('images').child(uuid)
+    await imageRef.put(blob, metadata)
+
+    return uuid
+  } catch (error) {
+    throw new Error('Storage Error: problem at image upload')
```
abelosorio: Same here.

abelosorio (approved):
EXCELLENT job! I've left a few comments. About yours, let me answer between lines:

> When you request a diagnose the image is stored in the bucket with user metadata (UID). Storage ID is created with UUIDv1. User can submit a diagnose without text, but not without an image. Should I change this?

No, that's fine. But let's use UUID v4 instead. They're more random.

> A diagnose is created in diagnose/ collection in firestore, containing text, images references and user. (weirdly It did not seem to contain timestamps). Should I add them?

Yes, let's agree on having `createdAt` and `updatedAt` fields for all records in the database.

> A diagnose reference is added to the user's diagnose array.

Perfect

> General project structure was improved. I also added an authenticationService for better separation of concerns, although I did not modify SignUp and Login, It's not this pull intent and it would be better if done in a separate one.

Perfect

:+1:

functions/index.js
```diff
@@ -0,0 +1,58 @@
...
+const admin = require('firebase-admin')
+const functions = require('firebase-functions')
+
+exports.onDiagnoseCreationAddTimestamp = functions.firestore
+  .document('diagnoses/{diagnoseId}')
+  .onCreate((snapshot, context) => {
+    const timestamp = admin.firestore.FieldValue.serverTimestamp()
+    return snapshot.ref.update({
+      createdAt: timestamp,
+      updatedAt: timestamp
```
abelosorio: The `updatedAt` value should only be set on updates. It should be null after the initial creation.

functions/index.js
```diff
@@ -0,0 +1,58 @@
...
+    })
+  })
+
+exports.onDiagnoseUpdateModifyTimestamp = functions.firestore
```
abelosorio: What does this function do?

functions/index.js
```diff
@@ -0,0 +1,58 @@
...
+    if (Math.abs(previousTimestamp.toMillis() - currentTimestamp.toMillis()) > FIVE_MINUTES_IN_MILLISECONDS) {
+      return snapshot.after.ref.update({
+        updatedAt: currentTimestamp
+      })
+    }
+
+    return null
+  })
+
+exports.onUserCreationAddTimestamp = functions.firestore
```
abelosorio: We can refactor this and create a single function to do this. Something as `setTimestampsToRecord('users/{userId}')`

abelosorio: cc @nahudalla

abelosorio (changes requested):

functions/index.js
MPucheta: I left it for future modifications on Diagnoses (such as submitting answers inside a diagnose).

src/authenticationService/index.js
```diff
@@ -0,0 +1,14 @@
...
+  try {
+    return firebase.auth().currentUser.uid
+  } catch (error) {
+    throw new AuthenticationError('Not logged in.')
+  }
+}
+
+export default {
+  getCurrentUserUID
+}
```
nahudalla: You should prefer exporting the individual elements rather than an object with al the exports:

```
export const getCurrentUserUID = ( ) => {
  // ...
}
```

Then you can import with:
```
import { getCurrentUserUID } from './path/to/file'
import * as allExportsInAnObject from './path/to/file'

console.log(getCurrentUserUID === allExportsInAnObject.getCurrentUserUID) // prints true
```

This also applies to other files.

src/imageService/index.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+const showImagePickerPromiseWrapper = (options) => (
+  new Promise((resolve, reject) => {
+    ImagePicker.showImagePicker(options, (imageResponse) => {
+      if (!imageResponse.uri) {
+        reject(new CameraError('Image not taken nor selected.'))
+      }
+      resolve(buildImageRepresentation(imageResponse))
```
nahudalla: Careful! You need to return in the `if` branch or add an `else`. Otherwise, you'll call both `reject` and `resolve` in case the `imageResponse` doesn't contain an `uri`.

src/screens/MainApp/DiagnoseRequest/DiagnoseSubmit/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { TouchableImage } from '~/helpers/TouchableImage'
+import styles from './styles'
+
+const DiagnoseSubmit = ({ onPress }) => (
+  <TouchableImage
+    style={styles.diagnoseSubmit}
+    onPress={onPress}
+    source={require('./resources/diagnoseSubmit.png')}
```
nahudalla: Can you use `import` for this? I think you should be able to do it. What is the result of this `require`? In most bundlers, this returns the path where the image is located (i.e. a string). In that case you can `import imagePath from './path/to/image'`. The `import` / `require` is usually needed because bundlers change the location/names of the files.

src/screens/MainApp/DiagnoseRequest/index.js
```diff
@@ -1,13 +1,79 @@
...
+import DatabaseService from '~/databaseService'
+import FullScreenActivityIndicator from '~/helpers/FullScreenActivityIndicator'
+import DiagnoseProblemDescription from './DiagnoseProblemDescription'
+import DiagnosePhoto from './DiagnosePhoto'
+import DiagnoseSubmit from './DiagnoseSubmit'
+import styles from './styles'
+import DiagnoseError from './DiagnoseError'
 
 const DiagnoseRequest = () => {
+  const baseImageSource = require('./resources/cameraIcon.png')
```
nahudalla: Same here about this `require`, can it be replaced by an `import`?

nahudalla (changes requested):
Only some minor comments. Excellent work! 💪

src/authenticationService/index.js
MPucheta: I changed this regarding this pull's files only. There may be some services (database and navigation) that still have some export default with object, changing them here would be messy as it affects SignUp/Login files. Can I do a separate pull for that only? (the same applies for require)

I left it as `import * as allExportsInAnObject from './path/to/file'`.

Otherwise it turned very hard to grasp interactions. For example:
`const imageURI = ImageService.getURIFromImage(image)`
`const imageReference = await StorageService.uploadImageAndReturnReference(imageURI)`
`await DatabaseService.addDiagnose(imageReference, problemDescription)`

would be:
`const imageURI = getURIFromImage(image)`
`const imageReference = await uploadImageAndReturnReference(imageURI)`
`await addDiagnose(imageReference, problemDescription)`

Even with descriptive names it becomes difficult to read. Then, I would have to replace `addDiagnose` to `addDiagnoseToDatabase` or maybe `submitDiagnose`, but reading it fast you are left wondering where it's submitted. Also without the `Service` part it gets difficult to see the separation of concerns. Let me know what you think!

nahudalla: > I changed this regarding this pull's files only. There may be some services (database and navigation) that still have some export default with object, changing them here would be messy as it affects SignUp/Login files. Can I do a separate pull for that only? (the same applies for require)

You can leave it as-is for now, it can be changed in a future iteration.

> I left it as `import * as allExportsInAnObject from './path/to/file'`.

I agree, it's not a bad approach. What I didn't like was exporting an object by default because it removes the possibility to import each export individually.

> Even with descriptive names it becomes difficult to read. Then, I would have to replace addDiagnose to addDiagnoseToDatabase or maybe submitDiagnose, but reading it fast you are left wondering where it's submitted. Also without the Service part it gets difficult to see the separation of concerns.

The names can be changed with the `import` statement too. If I have a file where I only need one of your exports and I want a more descriptive name, I could write:
```javascript
import {
  uploadImageAndReturnReference as uploadImageToStorage
} from './path/to/file/about/storage'
```

Although the names of the functions should be as descriptive as possible, I don't know about the `AndReturnReference` in the example above. I'd almost always expect a CRUD operation to return a reference to the created/updated object. Again, you can leave it as-is for now. No need to change the code right now.

nahudalla (approved):

abelosorio (approved):
Looks good!

nahudalla:
@MPucheta I saw you pushed new commits, configuring the APK build I think. Is this ready to merge? Please try to make smaller PRs with a clearly defined goal and freeze them once you achieve the goal so that we can safely merge them.

nahudalla (approved):

PR #6, opened 2019-10-25
[MP/6][ADDED] added instructions for building the project.

MPucheta:
- Downloading special configs folder from drive changes .env to _env, hence, I gitignored it just to be sure.

- I tested this from scratch and it worked. But I would prefer this to be merged after Jeremias's try.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /ios
 .env
+_env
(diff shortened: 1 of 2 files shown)


abelosorio (approved):
Looks good! ;)

PR #7, opened 2019-10-25
[HOTFIX.MP/5][CHANGED] changed image references to be an array

MPucheta:
- Changed image references to be treated as an Array, instead of string. 

- Changed add diagnose to add it's ID instead of reference. It turns out firebase is a mess to work with pure references.


abelosorio (approved):

abelosorio (approved):
Good design = code easy to change ;)

nahudalla (approved):

PR #8, opened 2019-10-26
[MP/7] Diagnose response screen implementation

MPucheta:
- Added a storage function to get URLs. 

- Added a navigation function to force re-renders. Tab navigation does not un-mount components, so to get new data upon entering the response screen, we need a way to cause a refetch.

- Added a database function to fetch answers. It doesn't have a cache or something like that, so it fetches up to 25 documents (howMany parameter) from the user, ordered by 'updatedAt' so new answers are placed at the top level (older ones appear too, and you can scroll-down to see them). Firestore do not support queries for null, so I query for the diagnoses for the user and manually check if that field exists.

- Although it's good for now to implement this using the updatedAt, as the only way to update a diagnose is to put an answer. I think we could have an answeredAt field to be more explicit and as a guard should another way to update a diagnose appears.

- As I'm exploring firebase limitations and such I'm worried about database schema. Although it's a noSQL, we should decide the data types as those conditions how we think the logic of the functions. For example, I was tempted to model answers as an Array instead of a String (obviously, I can just implement the function to check/cast to array, but I do not think it would be good to clutter the code with checks for something that could be static). 

- There is a composite index for (user, updatedAt) to easily execute queries. It should be used in production.

**This uses the HOTFIX (image as array and storing diagnoses by id instead of references**

screencast-diagnose-response


abelosorio (dismissed):

src/screens/MainApp/DiagnoseResponse/index.js
```diff
@@ -1,12 +1,69 @@
...
+    // Tab navigator doesn't unmount components, we need this to refetch need data when pressing DiagnoseResponse tab.
+    // alternative is to add a query snapshot listener, so that new changes in the database trigger a re-render automatically.
+    toggleRefetch(!refetchTrigger)
+  }
+
+  const buildDiagnoses = async () => {
+    setDownloadingDiagnoses(true)
+    const answers = await getDiagnosesFromDatabase()
+    const diagnoses = []
+    for (let i = 0; i < answers.length; i++) {
```
abelosorio: Consider using Array's methods instead.

abelosorio (commented):
Looks good! I agree with @nahudalla that PRs should be smaller, it's easier for those like me that are not that into the project's details to review.
Also, it could be great to see a screencast of the new functionality it adds.

Thanks!

src/databaseService/index.js
```diff
@@ -81,14 +81,45 @@ const addDiagnoseIDToCurrentUser = async (diagnoseID, collectionPath = 'users')
...
+    querySnapshot.forEach((doc) => {
+      const data = doc.data()
+      if (data.answer) {
+        answers.push({
+          imageReference: data.imageReferences[0],
+          answer: data.answer,
+          answeredBy: data.answeredBy
+        })
+      }
+    })
```
nahudalla: This can be done with `map` and `filter`. I think it's always easier to read and reason about with array methods. Also, you could add an `answered` field with a boolean value to filter requests that have answers on the Firebase side.

How to do the same with `map` and `filter`:
```javascript
return querySnapshot.docs
  .map(doc => doc.data())
  .filter(doc => doc.answer)
  .map(doc => ({ ...doc, imageReference: doc.imageReference[0] }))
```

src/screens/MainApp/DiagnoseResponse/index.js
MPucheta: Is this regarding forEach? I run into a weird problem using async/await inside forEach. Like it didn't execute due to each forEach function being a Promise.

I tried something like:

`answers.forEach(async(answer)=>{...})`

But I never specify that I should wait for the functions inside the forEach to finish. Do you suggest I use something like the following:

```
Promise.all(
answers.forEach(...)
)
```

This doesn't preserve ordering though.

**Note: I added the screencast too.**

src/screens/MainApp/DiagnoseResponse/Diagnose/index.js
```diff
@@ -0,0 +1,34 @@
...
+  thumbnail
+    ? <Image
+      source={{ uri: thumbnail }}
+      style={styles.thumbnail}
+    />
+    : <Image
+      source={cannabisLeaf}
+      style={styles.thumbnail}
+    />
+)
```
nahudalla: Please apply the ternary conditional on the prop directly:
```
<Image
  source={thumbnail ? { uri: thumbnail } : cannabisLeaf}
  style={styles.thumbnail}
/>
```

nahudalla (changes requested):
:muscle: Good work!

src/screens/MainApp/DiagnoseResponse/index.js
nahudalla: Yes, it can be done with Array methods and `Promise.all`, but you need to use `map` instead of `forEach`. Also keep in mind you'll be doing operations in parallel. You _can_ force the async operations to be run in sequence with `reduce`. The current behavior with your `for` loop is to run them in sequence. Example:
```javascript
// in parallel
const diagnoses = await Promise.all(answers.map(async answer => {
  // some async operation to return the diagnose
}))

// in sequence
const diagnoses = await answers.reduce(
  async (diagnoses, answer) => {
    // diagnoses will be a promise that resolves to the result of the previous reduction
    diagnoses = await diagnoses // wait for the previous diagnose to finish
   
    const diagnose = await somehowGetTheDiagnoseAsynchronously() // get the new diagnose

    diagnoses.push(diagnose) // push the diagnose to the diagnoses array

    return diagnoses // return the diagnoses array with the new diagnose
  },
  Promise.resolve([])
)
```

abelosorio (approved):

nahudalla (approved):

PR #9, opened 2019-10-29
[MP] Diagnose response input web interface

MPucheta:
- Implemented web interface. I decided to keep this code separated from the mobile one. Its a little code repetition vs better separation of concerns, also changes in the mobile files won't affect web files.

- I did not try to deploy this using firebase hosting and, as a consequence, I do not upload public/. Also .cache/ and dist/ were heavy, so I added them to .gitignore.

- A little CSS was necessary, as the interface was too clunky to manage.

- Fetches diagnoses by the answered field.

- This is only the response feature, I consider the notification to the user a separate feature.

- I left support for multiple images, as imageReferences is determined to be an array. I manually copied Abel's cactus picture to the diagnose's image array.

- **NEW CHANGES** Fetching diagnoses now takes into consideration multiple users accessing the page, and also if someone else writes an answer. A new screencast was added to show it.

screencast-diagnose-response-V2

image

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   presets: [
     'module:metro-react-native-babel-preset',
-    'module:react-native-dotenv',
+    'module:react-native-dotenv'
   ],
   plugins: [
     [
--- a/babel.config.js
+++ b/babel.config.js
@@ -2,11 +2,6 @@ module.exports = {
   presets: [
     'module:metro-react-native-babel-preset',
     'module:react-native-dotenv',
-    '@babel/preset-env', {
-      targets: {
-        node: 'current'
-      }
-    }
   ],
   plugins: [
     [
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,7 +1,12 @@
 module.exports = {
   presets: [
     'module:metro-react-native-babel-preset',
-    'module:react-native-dotenv'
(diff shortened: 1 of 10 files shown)


abelosorio (approved):

babel.config.js
```diff
@@ -1,7 +1,12 @@
 module.exports = {
   presets: [
     'module:metro-react-native-babel-preset',
-    'module:react-native-dotenv'
+    'module:react-native-dotenv',
+    '@babel/preset-env', {
+      targets: {
+        node: 'current'
+      }
+    }
```
nahudalla: Be careful, you're compiling for a Browser, not `node`. If you did this to enable support of `async/await` (`regeneratorRuntime undefined` error), you should only add the following to `package.json`. ParcelJS will then compile the bundle with Promises support:
```JSON
"browserslist": [
  "last 1 Chrome versions"
]
```

nahudalla (commented):
:muscle: Great work! Only one comment below.

Also let's use as minimum CSS as possible for now. If we need to build this "admin" interface further, we'll be using `SASS` with `PostCSS Modules` for compiling `CSS` in the future. `Bootstrap`'s CSS will also be imported through a `SASS` import if it's needed. ParcelJS does all that automatically with almost no configuration, but let's leave that for later.

nahudalla (approved):

PR #10, opened 2019-10-31
This is the first implementation of the Dashboard

jmbidev:
To implement it I added a function in databaseService/index.js and modified Home/index.js replacing the old content with the dashboard. I also moved the logout button in home to Settings/index.js.

Screenshot_1572545630


src/databaseService/index.js
```diff
@@ -114,6 +114,22 @@ export const getAnsweredDiagnosesForCurrentUser = async (howMany = 25, diagnoseP
...
+  try {
+    const currentUserUID = AuthenticationService.getCurrentUserUID()
+    const querySnapshot = await firebase
+      .firestore()
+      .collection(diagnosePath)
+      .where('user', '==', currentUserUID)
+      .limit(howMany)
+      .get()
+
+    return querySnapshot.docs.map(doc => ({ ...doc.data() }))
```
nahudalla: This object spread is not necessary: `({ ...doc.data() })` is the same as `doc.data()`

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/BoardItem/index.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
```
nahudalla, line 1: I don't know if this was something already present, but I don't like the way these components are nested one inside the other (`MainApp/Home/Dashboard/DiagnosesBoard/BoardItem`). As a general rule, there should never be more than one level of nesting in the components.

As I understand it, `Home` is the same as `Dashboard` here. `Dashboard` should be a _page_ (or _view_ in native apps terminology) inside the `MainApp` module. `DiagnosesBoard` should be a _component_ inside the `MainApp` module. `BoardItem` can stay inside the `DiagnosesBoard` component (one level of nesting).

So the structure would be:
- `modules/MainApp/views/Dashboard/index.js`: the entry point for the dashboard.
- `modules/MainApp/components/DiagnosesBoard/index.js`: the entry point for the `DiagnosesBoard` component.

cc @MPucheta , this is something we need to discuss. No need to change it in this PR.

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/BoardItem/index.js
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const AnsweredIcon = ({ answered }) => {
+  const icon = (answered === false) ? notAnsweredIcon : answeredIcon
+  const style = (answered === false) ? styles.notAnswered : styles.answered
+  return (<Image source={icon} style={style} />)
+}
+
+const BoardItem = (props) => {
+  const date = new Date(props.diagnose.createdAt.seconds * 1000)
```
nahudalla: You can use a library such as `moment.js` for dealing with dates and times.

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+const buildBoardItems = async (downloadedData) => (
+  downloadedData.reduce(async (boardItems, diagnose, index) => {
+    boardItems = await boardItems
+    const boardItem = await getBoardItemFromDiagnose(diagnose, index)
+    boardItems.push(boardItem)
+
+    return boardItems
+  },
+  Promise.resolve([])))
```
nahudalla: Wrong indentation here. It should be:
```javascript
const buildBoardItems = async (downloadedData) => (
  downloadedData.reduce(
    async (boardItems, diagnose, index) => {
      boardItems = await boardItems
      const boardItem = await getBoardItemFromDiagnose(diagnose, index)
      boardItems.push(boardItem)

      return boardItems
    },
    Promise.resolve([])
  )
)
```

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/styles.js
```diff
@@ -0,0 +1,24 @@
...
+    height: 44,
+    textAlign: 'left',
+    textAlignVertical: 'center'
+  },
+  listContainer: {
+    flex: 1,
+    overflow: 'hidden'
+  },
+  list: {
+  }
```
nahudalla: Why this empty class?

src/screens/MainApp/Home/Dashboard/NewDiagnoseBoard/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { View, Text, TouchableOpacity } from 'react-native'
+import styles from './styles'
+
+const NewDiagBoard = (props) => {
```
nahudalla: Please don't use abbreviations on names. Why the word `New` here?

src/screens/MainApp/Home/Dashboard/NewDiagnoseBoard/index.js
```diff
@@ -0,0 +1,28 @@
...
+        </Text>
+        <Text style={styles.text}>
+          ¿Parece estar enferma?
+        </Text>
+        <Text style={styles.text}>
+          No te preocupes más
+        </Text>
+      </View>
+      <TouchableOpacity
+        onPress={() => { props.handleRedirectFunc() }}
```
nahudalla: - Functions can be passed directly as event handlers. No need to create a lambda: `onPress={handleRedirectFunc}`.
- Prefer object destructuring to receive props: `const NewDiagBoard = ({ handleRedirectFunc }) => { ... }`.
- A better naming for this prop would be `redirectHandler`. Although I prefer more descriptive names such as `goToCreateDiagnoseScreen`.

nahudalla (changes requested):
Great work as your first PR! Pay attention to the following:
- Naming. You should prefer more descriptive names and avoid abbreviations.
- Newer JavaScript features (such as object destructuring).
- Use libraries that implement what you need (such as `moment.js` to handle dates).
- Avoid copy/pasting code as much as you can. Always try to find a way to reuse existent code.
- Avoid nesting components inside other components.
- We'll work on refactoring at a later time. No need to make big changes now (such as changing the components structure).

cc @MPucheta :point_up_2: some of these comments apply to you too.

src/screens/MainApp/Home/Dashboard/NewDiagnoseBoard/index.js
jmbidev: I used the 'new' word because that board is supossed to redirect to make a new diagnose request

nahudalla: Oh I see. Maybe a more descriptive name could be `CreateDiagnoseBanner`?

src/screens/MainApp/Home/Dashboard/NewDiagnoseBoard/index.js
jmbidev: The standard js plugin I'm using in visual studio code says that function handlers' name should start with 'handle' word. So it should be onPress={handleSomething} according to it. Suggestion?

src/screens/MainApp/Home/Dashboard/NewDiagnoseBoard/index.js
jmbidev: I renamed it to CreateDiagnoseBoard

src/screens/MainApp/Home/Dashboard/NewDiagnoseBoard/index.js
nahudalla: :thinking: This is the first time I see that suggestion from StandardJS. I think the problem here is about the component itself. I'm not sure if it deserves to be a component. Anyway, how about `handleGoToCreateDiagnoseScreen`?

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/BoardItem/index.js
```diff
@@ -0,0 +1,48 @@
...
+    source={thumbnail ? { uri: thumbnail } : cannabisLeaf}
+    style={styles.thumbnail}
+  />
+)
+
+const AnsweredIcon = ({ answered }) => {
+  const icon = (answered === false) ? notAnsweredIcon : answeredIcon
+  const style = (answered === false) ? styles.notAnswered : styles.answered
+  return (<Image source={icon} style={style} />)
+}
```
nahudalla: - No need to test if it equals `false`. The following is more readable:
```javascript
  const icon = answered ? answeredIcon : notAnsweredIcon
  const style = answered ? styles.answered : styles.notAnswered
```
- No need to use parenthesis on one-line return statements.

nahudalla (changes requested):

src/screens/MainApp/Home/Dashboard/NewDiagnoseBoard/index.js
jmbidev: As you said maybe the component itself is the problem. I made it that way because I thought that dashboard could be implemented as an union of independents board components with different functionality.

jmbidev: I left it as handleGoToCreateDiagnoseScreen

src/databaseService/index.js
```diff
@@ -108,7 +108,23 @@ export const getAnsweredDiagnosesForCurrentUser = async (howMany = 25, diagnoseP
       .limit(howMany)
       .get()
 
-    return querySnapshot.docs.map(doc => ({ ...doc.data() }))
```
MPucheta, line 118: I do not see this function being used in the pull, It could still we working after this change, but it seems more like a refactoring than a functionality. Be careful when doing these things, even if something worked for a similar case, you could still be introducing bugs. As an instance, doc.data() returns a DocumentData, which may or may not be as accessible as a plain object.

When developing you typically have two hats, a "funcionality hat" and a "refactoring hat". A different mindset applies when doing one or another, and it tipically involves having many unit tests which we are lacking.

src/screens/MainApp/Home/Dashboard/CreateDiagnoseBoard/index.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { View, Text, TouchableOpacity } from 'react-native'
+import styles from './styles'
+
+const CreateDiagnoseBoard = ({ handleGoToCreateDiagnoseScreen }) => {
+  return (
+    <View style={styles.container}>
+      <View style={styles.title}>
+        <Text style={styles.text}>
```
MPucheta: I created a reusable component in helpers "AppText" which has a consistent style (roboto 14). It also has the property of being customizable, so you could only add the font size 18 adhoc.

It's something minor but you could save a little verbosity from applying styles. We can also enforce some consistency.

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/BoardItem/index.js
```diff
@@ -0,0 +1,47 @@
...
+)
+
+const AnsweredIcon = ({ answered }) => {
+  const icon = answered ? answeredIcon : notAnsweredIcon
+  const style = answered ? styles.answered : styles.notAnswered
+  return <Image source={icon} style={style} />
+}
+
+const BoardItem = ({ thumbnail, diagnose }) => {
+  const moment = require('moment')
```
MPucheta: Please consider moving `require` to the top. Also, we use `import`.
Something like this:
`import { View, Text, Image } from 'react-native'`
`import moment from 'moment'`

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/index.js
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <BoardItem
+      key={key}
+      thumbnail={thumbnail}
+      diagnose={diagnose}
+    />
+  )
+}
+
+export const sortByCreatedAt = (diagnoses) => (
```
MPucheta: Careful when using previous code. This `export` is from a utility module.

src/databaseService/index.js
jmbidev: The reason I changed it... https://github.com/WeDevelop-ARG/cannabis-app/pull/10#discussion_r341626483

nahudalla: What you say is very true, Mauro.

I asked him for this change because I knew it'd work: Everything in JavaScript is an object (well, almost everything). `doc.data()` returns an object, whose prototype chain contains `DocumentData`. Any object in JavaScript have _enumerable_ and _non-enumerable_ properties. The object spread syntax (`{ ...obj }`) is just syntactic sugar to copy all _enumerable_ properties from one object to the other. When you use spread, you're actually **not copying** some properties from the original object, the properties that are _non-enumerable_.

In cases like the Firebase Library, they only tell you that the object is an instance of certain class (`DocumentData`), but they don't give information on how the properties are defined on that object. So, to be safe, it's best to avoid copying the properties here. Other libraries (such as Sequelize) contain methods to obtain a plain JavaScript object from their custom one, so that you know what you're working with and get the expected results.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

src/screens/MainApp/Home/Dashboard/DiagnosesBoard/index.js
jmbidev: It would be nice to have that function in a different place since we both used it.

nahudalla (approved):

abelosorio (approved):
Looks good. @jmbidev please fix the conflicts when you have a chance.

MPucheta (approved):

PR #11, opened 2019-11-01
[MP] Added metrics with firebase SDK integration and react-native-firebase library

MPucheta:
- I had to add react-native-firebase library to the project, it introduces quite a few breaking changes. I would prefer If we stick to Firebase JS SDK for the things that work (firestore, storage, and such) and only rely on react-native-firebase on things that do not work (analytics, cloud messaging, etc). Although, If it's necessary, it shouldn't be **that** hard to migrate (majority of firebase is encapsulated in each service, so changing the implementation could be just one line in each service file).

- As now, it's logged when a user submits a diagnose, and when signing up.

- I attach the debug view, logged events are expected to appear in the next 24hs (as it says in the firebase console).

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ gradle.properties
 build/
 .cache/
 dist/
+google-services.json
+.firebaserc
 
 #All next is react-native .gitignore
 
--- a/android/app/BUCK
+++ b/android/app/BUCK
@@ -35,12 +35,12 @@ android_library(
 
 android_build_config(
     name = "build_config",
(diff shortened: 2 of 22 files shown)


nahudalla (commented):
Great work overall. Waiting for the package name change before approval.

I agree we should stick to the official Firebase library for things that work with React Native and use `react-native-firebase` for things that don't.

nahudalla (approved):

jmbidev (approved):

abelosorio (approved):
Looks great!

PR #12, opened 2019-11-02
[MP] Added tab icons. Changed App icons and display name.

MPucheta:
- Added tab icons with active and inactive colors. The library, react-native-vector-icons has a lot of other icons should we need them (for google, facebook, and such).

image

image


android/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,3 @@
 <resources>
-    <string name="app_name">CannabisApp</string>
+    <string name="app_name">DrCannabis</string>
```
abelosorio, line 2: Can we rename it to `Dr. Cannabis`?

abelosorio (approved):

nahudalla (approved):

android/app/src/main/res/values/strings.xml
nahudalla: It should be easy to change. @MPucheta ?

PR #19, opened 2019-11-21
[MP] Social Network Authentication

MPucheta:
- Added Google Sign in method, this requires a new environmental variable and changes in google-services. SIgnin is done after user puts a username, otherwise the user could signin with google, restart the application before putting a username, and still use the app.

- Implemented using Marcos's proposed flow and a placeholder styling. Only thing different is that I did not implement the function that suggests a new nickname if the one that was inserted was already taken (should I make nicknames at random, take the inserted one and append random numbers to them?)

- If the user is new and logs with google, then he has to put a new username

- If the user already had an account with that email and it was not verified, then upon using google signin the username is reused BUT he cannot longer use email authentication. **This is the expected functionality described by the firebase developers, the google signin authentication OVERWRITES the unverified email authentication**

- Changed LoadingScreen to unsuscribe the observer on authentication changes. It's only necessary on app startup.

- Renamed old Login to LoginEmail. New Login is the screen with the google button.

image
image (1)


src/databaseService/index.js
```diff
@@ -65,6 +65,18 @@ export const queryEmailFromUsername = async (username, collectionPath = 'users')
...
 }
 
+export const queryUsernameFromEmail = async (email, collectionPath = 'users') => {
+  try {
+    const querySnapshot = await firebase.firestore().collection(collectionPath).where('email', '==', email).get()
+    if (hasDocuments(querySnapshot)) {
+      const json = getFirstDocumentDataFromQuerySnapshot(querySnapshot)
+      return json && json.username
+    }
+  } catch (error) {
```
nahudalla, line 75: Please add an empty line after variable declarations (`const querySnapshot`) and the next line. Also, add an empty line before `return` statements.

nahudalla (approved):

jmbidev (approved):

PR #20, opened 2019-11-22
[MP] Adding-Splash-Screen

MPucheta:
- I was tempted to include a transition or animation, but choose not to as it was not specified. If you would like a fade in for the screen or something like that, let me know and I would be happy to implement it.

- **Also, I would really prefer if this was merged after the previous pull, just to maintain order and because some similarities (assets folder, react-native-size-matters)**

DrCannabis


abelosorio (approved):

nahudalla (approved):

src/screens/LoadingScreen/index.js
```diff
@@ -26,11 +31,20 @@ const LoadingScreen = () => {
         checkIfLoggedIn()
       }
     }
-    decideIfGoToPrivacyPolicyScreenOrLoginScreen()
+
+    setTimeout(() => decideIfGoToPrivacyPolicyScreenOrLoginScreen(), 4000)
```
jmbidev, line 35: Just a doubt. Are those 4 seconds the duration of the loading screen? Maybe could be customizable instead of fixed value?

jmbidev (approved):

src/screens/LoadingScreen/index.js
nahudalla: My question is, do we really need those 4 seconds? I think the faster the app loads, the better. Even if the screen is not shown at all because no loading happens, it's better.

@MPucheta

nahudalla:
@MPucheta please resolve the conflicts

PR #21, opened 2019-12-02
[JB][Diagnose response push notifications]

jmbidev:
Added push notifications using notification groups to manage user's devices.

Considerations:

- We need to add firebase functions configuration to make it run. This can be done through firebase cli with the commands:
 `firebase functions:config:set fcm.api_key=<api_key>`
 `firebase functions:config:set fcm.sender_id=<sender_id>`

- Notifications only shows up when the app is running in background. To show notifications in foreground we should use @react-native-firebase/notifications module which isn't currently supported in version 6 of react-native-firebase. 

- FCM Tokens are stored in notification groups with FCM api, so don't need to store them in the project's database

- Added node-fetch dependency to manage http requests to FCM api

Screenshot_1575302765


functions/index.js
```diff
@@ -43,5 +44,118 @@ const onAnswerMigrateToUserCollection = functions
...
+        'Content-Type': 'application/json',
+        Authorization: `key = ${functions.config().fcm.api_key}`,
+        project_id: functions.config().fcm.sender_id
+      }
+    }
+    if (method === 'post') {
+      options.body = JSON.stringify(body)
+    }
+    const response = await fetch(url, options)
+    return await response.text()
```
MPucheta, line 90: Please add a blank line before the return.

In general, logical groupings should be separated by blanks lines. For example, this if clause should have blanks before and after

functions/index.js
```diff
@@ -43,5 +44,118 @@ const onAnswerMigrateToUserCollection = functions
...
+  } catch (error) {
+    console.log(error)
+    return null
+  }
+}
+
+const getNotificationKey = async (notificationKeyName) => {
+  const url = `https://fcm.googleapis.com/fcm/notification?notification_key_name=${notificationKeyName}`
```
MPucheta, line 99: Add a blank line after variable declarations.

src/messagingService/index.js
```diff
@@ -0,0 +1,45 @@
+import messaging from '@react-native-firebase/messaging'
+import firebase from 'firebase'
+import 'firebase/functions'
+
+export const checkForPermissions = async () => {
+  const enabled = await messaging().hasPermission()
+  if (!enabled) {
+    try {
+      await messaging().requestPermission()
+    } catch (error) {
```
MPucheta, line 13: Mmm maybe I would expect this to throw MessagingError or at least console.log(error.message). The same applies for other functions in this file.

functions/index.js
```diff
@@ -43,5 +44,118 @@ const onAnswerMigrateToUserCollection = functions
...
+  .onCall(async (data, context) => {
+    const fcmToken = data.fcmToken
+    const uid = context.auth.uid
+    const notificationKeyName = 'appUser-' + uid
+    const notificationKey = await getNotificationKey(notificationKeyName)
+
+    if (notificationKey !== null) {
+      await runFCMOperation('remove', notificationKeyName, notificationKey, [fcmToken])
+    }
+  })
```
MPucheta, line 163: i'm having trouble seeing the execution flow of this. On a https call, those two functions will fire (in "parallel"). 
On `notificationKey !== null`:

-  `storeFCMTokenInUserGroup` will do `runFCMOperation('add')`, and 

-  `removeFCMTokenInUserGroup` will do `runFCMOperation('remove')`.

Can you elaborate a little bit the expected behaviour (in this discussion)?

functions/index.js
```diff
@@ -43,5 +44,118 @@ const onAnswerMigrateToUserCollection = functions
...
+  return null
+}
+
+const createNotificationGroup = async (notificationKeyName, fcmToken) => {
+  const url = 'https://fcm.googleapis.com/fcm/notification'
+  const data = {
+    operation: 'create',
+    notification_key_name: notificationKeyName,
+    registration_ids: [fcmToken]
+  }
```
MPucheta, line 120: Here too, a blank after variable declarations.

functions/index.js
```diff
@@ -43,5 +44,118 @@ const onAnswerMigrateToUserCollection = functions
...
+}
+
+const runFCMOperation = async (operation, notificationKeyName, notificationKey, fcmTokens) => {
+  const url = 'https://fcm.googleapis.com/fcm/notification'
+  const body = {
+    operation: operation,
+    notification_key_name: notificationKeyName,
+    notification_key: notificationKey,
+    registration_ids: fcmTokens
+  }
```
MPucheta, line 132: Here too, a blank after variable declarations.

src/messagingService/index.js
```diff
@@ -0,0 +1,45 @@
+import messaging from '@react-native-firebase/messaging'
+import firebase from 'firebase'
+import 'firebase/functions'
+
+export const checkForPermissions = async () => {
+  const enabled = await messaging().hasPermission()
```
MPucheta, line 8: A blank after variable declarations.

MPucheta (changes requested):
Great Work!

I left some comments, mostly on code conventions (blank lines, logical groupings, etc). I also would like some insight in the thought process behind the functions `storeFCMTokenInUserGroup` and `removeFCMTokenInUserGroup`

functions/index.js
abelosorio: The only exception is when the command is the only one within a block, for example:
```
if (condition) {
  return value
}
```

abelosorio (approved):
Looks good

functions/index.js
jmbidev: These functions works a little bit different to the ones that listens to database changes. 
These functions are callables from the client by the function's name.

From MessagingService module we can call them like this...

-`
const removeFCMTokenInUserGroup = firebase.functions().httpsCallable('removeFCMTokenInUserGroup')`
`await removeFCMTokenInUserGroup({ fcmToken: fcmToken })`

-`const storeFCMTokenInUserGroup = firebase.functions().httpsCallable('storeFCMTokenInUserGroup')`
`await storeFCMTokenInUserGroup({ fcmToken: fcmToken })`



- `storeFCMTokenInUserGroup` is called when users **login**, it's purpose is to save the fcm token into a existing user's devices group, or into a new user's devices group if there isn't one.

- `removeFCMTokenInUserGroup` is called when users **logout**, it's purpose is to remove the fcm token in a existing user's devices group. FCM api automatically deletes empty groups that's why we need to create them again when the user login.

src/messagingService/index.js
jmbidev: I initially thrown errors but because of how are called from login, I removed those lines.
It could make the user login fail after user signed in, so I think it should be console log until we rework login screen

jmbidev: If the user doesn't give permissions, the app should keep going on. It will only affect notifications

MPucheta (approved):

functions/index.js
MPucheta: Amazing! I shall write this as a recipe.

src/screens/MainApp/Settings/index.js
```diff
@@ -4,9 +4,11 @@ import NavigationService from '~/navigationService'
 import * as firebase from 'firebase'
 import * as AnalyticsService from '~/analyticsService'
 import { goToPrivacyPolicyURL } from '~/mixins/privacyPolicyMixins'
+import MessagingService from '~/messagingService'
 
 const LogOut = async () => {
   try {
+    await MessagingService.deleteFCMTokenForCurrentUser()
     await firebase.auth().signOut()
     NavigationService.navigate('Login')
```
nahudalla, line 13: If the FCM Token deletion fails, the user won't be signed out. I think it should be in a separate try/catch block, since we can safely ignore that error and continue with the logout flow.

src/screens/authentication/Login/index.js
```diff
@@ -20,11 +21,15 @@ const onGoogleButtonPress = async () => {
     const userEmail = signInData.user.email
     if (await AuthenticationService.emailAlreadyHasGoogleSignIn(userEmail)) {
       await AuthenticationService.loginWithGoogleSignInData(signInData)
+      await MessagingService.checkForPermissions()
+      await MessagingService.saveFCMTokenForCurrentUser()
```
nahudalla: Do any of these functions throw an error? If they do, extract these two lines to a function and surround them with a try/catch, so that we can safely ignore the error and the login flow continues as normal.

src/screens/authentication/Login/index.js
```diff
@@ -20,11 +21,15 @@ const onGoogleButtonPress = async () => {
...
       NavigationService.navigate('MainApp')
     } else {
       const username = await DatabaseService.queryUsernameFromEmail(userEmail)
       if (username) {
         await AuthenticationService.loginWithGoogleSignInData()
+        await MessagingService.checkForPermissions()
+        await MessagingService.saveFCMTokenForCurrentUser()
```
nahudalla: Same here, also you're repeating code (calling these two lines in the same order). This should definitely be on its own function. Maybe `enableNotificationsForCurrentUser`.

src/screens/authentication/LoginEmail/index.js
```diff
@@ -32,8 +33,9 @@ const LoginEmail = () => {
       } else {
         email = values.account
       }
-
       await firebase.auth().signInWithEmailAndPassword(email, values.password)
+      await MessagingService.checkForPermissions()
+      await MessagingService.saveFCMTokenForCurrentUser()
```
nahudalla: Again, this should be in `enableNotificationsForCurrentUser`

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -51,6 +52,8 @@ const UsernameRequest = (props) => {
     } finally {
       setSubmitting(false)
       if (signedIn) {
+        await MessagingService.checkForPermissions()
+        await MessagingService.saveFCMTokenForCurrentUser()
```
nahudalla: Again.

nahudalla (changes requested):
Looks good! Just a few changes needed. You're violating the DRY principle with the code that enables notifications for a user.

Also, **please use `console.error` instead of `console.log` to report errors** in the console. The former uses the **standard _error_ output**, whereas the latter uses the **standard output** (not error). I didn't comment this on the code since it repeated on every try/catch.

src/screens/MainApp/Settings/index.js
jmbidev: That function doesn't throw any error. Same as the one checking for permissions...

src/screens/authentication/Login/index.js
jmbidev: Those functions doesn't throw errors. Also I talked this with Mauro because we believe that login will have a rework for the new wireframes where those functions will only be needed in 1 place...

MPucheta: Although it's true that some things will change for the v4 wireframe and the login process can be encapsulated, I second Nahuel here. This piece of information should only have one representation. 

Even if this ends up only needed in one place, then this block of `checkForPermissions` and `saveToken`would still be those two lines. If `saveToken` assumes that push notification were allowed, then it can sustain itself without `checkForPermissions`, but this is not the case as a previous check is always needed. Everytime a `saveToken` is done, it has to be preceded with a `check` (there is some implicit 'coupling' there). Although it's somewhat true that if it only appears in one place, then DRY is not violated (Debatable), but as it stands, Nahuel is right.

Let's decide which part of push notifications is mandatory to have.

-  If the user doesn't allow them so he can't login. If this is the case lets throw an error and check for this at the start of the login process.

- If the user doesn't allow them and he can login. If this is a case a thrown error shouldn't interfere with the login process (log the error or have a nested try/catch take care of that)

- If it doesn't matter if the token is deleted or not, then an error on that function shouldn't block the signing out (log the error or have a nested try/catch take care of that)

- If it does matter that the token is deleted or not, then an error on this function should block the signing out (I doubt this could result in an error unless external factors, but if it happens, then no action is done or a reusable modal for unexpected errors can be shown).

For consistency, can we decide to have all services throw an error and later, in the consumer code, decide if that error does or doesn't have to block the normal flow with nested try/catchs?

@nahudalla @jmbidev

jmbidev: Actually, it's not required to call checkForPermissions every time we call saveFCMTokenForCurrentUsers. It is a coincidence that I decided to use those two together. 

checkForPermissions will ask if messaging is enabled and if not will request those permissions.
saveToken will ask if there are permissions to proceed sending the token, but won't request.

I think the DRY was already violated with the redirections to MainApp, and even if we group the two functions in a single function still will be violated since It's in many place while it could be all in the same place. But I think that will be fixed with the new wireframes.

Btw that doesn't mean that those functions can't be on the same function, so I will group them. 

And I agree to throw erros in services for consistency

jmbidev: > - If the user doesn't allow them so he can't login. If this is the case lets throw an error and check for this at the start of the login process.
> - If the user doesn't allow them and he can login. If this is a case a thrown error shouldn't interfere with the login process (log the error or have a nested try/catch take care of that)

I think the second option is the correct.


> - If it doesn't matter if the token is deleted or not, then an error on that function shouldn't block the signing out (log the error or have a nested try/catch take care of that)
> - If it does matter that the token is deleted or not, then an error on this function should block the signing out (I doubt this could result in an error unless external factors, but if it happens, then no action is done or a reusable modal for unexpected errors can be shown).

I think this function should allways delete if there is a token set. And prevent user logout if couldn't remove the token. Because if it wasn't removed the device will still receive notifications for that user. So if you are ok with it I'll make the changes needed.

@nahudalla @MPucheta

src/messagingService/index.js
```diff
@@ -9,39 +11,45 @@ export const checkForPermissions = async () => {
...
-        await removeFCMTokenInUserGroup({ fcmToken: fcmToken })
-      } catch (error) {
-        console.log(error.message)
-      }
+    await CacheService.removeItem('fcmToken')
+  } catch (error) {
+    fcmToken = null
+  }
+
+  if (fcmToken) {
```
nahudalla, line 46: If I remember correctly, there's a rule in Clean Code that forbids doing this. You're basically checking if it didn't fail to continue with the rest of the code. You could've fetched the token, and do the rest only if a token exists:

```javascript
const fcmToken = await CacheService.getItem('fcmToken')

if (!fcmToken) return undefined

try {
  // remove the tokem from the cache and FCM group
} catch (err) {
  // throw an appropriate error
}
```

Don't change it now tho, we need to advance with new features

src/messagingService/index.js
```diff
@@ -9,39 +11,45 @@ export const checkForPermissions = async () => {
...
+    try {
+      const removeFCMTokenInUserGroup = firebase.functions().httpsCallable('removeFCMTokenInUserGroup')
+      
+      await removeFCMTokenInUserGroup({ fcmToken: fcmToken })
```
nahudalla, line 50: When you're creating an object that has a key with the same name of the variable that has the value being assigned, you can skip the value assignment: `await removeFCMTokenInUserGroup({ fcmToken })`

nahudalla (approved):
Great work! :muscle: No need to make these changes now, the comments are just informative.

PR #22, opened 2019-12-05
[MP] Adding onBoarding

MPucheta:
- Added onBoarding, already wireframed to SignUp and with a cached value to only show it the first time the app is launched.

- I did not change anything related to login or seeing if the user has username or something like that (I considered it yesterday but it was not really the responsibility of this branch)

- first, second, and third.png are just the logo as placeholder for the real images. I consider stealing something from google but it seemed better to leave it like that and just ask the designers for the real ones (if this is not a possibility at the time of deploy, then there we search for ones in google).

- Gif of the functionality (I manually set off the cache to show both paths that the onBoarding can be interacted)

onBoarding


src/App.js
```diff
@@ -13,6 +13,7 @@ import LoginEmail from './screens/authentication/LoginEmail'
 import UsernameRequest from './screens/authentication/UsernameRequest'
 import Home from './screens/MainApp/Home'
 import LoadingScreen from './screens/LoadingScreen'
+import OnBoarding from './screens/OnBoarding'
```
nahudalla: "onboarding" is a single word. This module should be called `Onboarding`

nahudalla (approved):

jmbidev (approved):
Looks nice!

nahudalla (approved):

PR #23, opened 2019-12-09
[MP/Adding-theme-support-and-reusable-components]

MPucheta:
- Added theme.js, which encapsulates colors, sizes and fonts.

- Added Text and Button generic components. They accept props for automatic styling based on theme, they are also customizable (you can manually apply new styles and overwrite the theme ones should we need that). This Text replaces AppText (this new one is more flexible).

- restProps pattern is used, so this Text and Button has the same interface as react-native Text and TouchableOpacity

Sample usage

image

Default props are those with more common usage. Proptypes added to enforce specific values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ node_modules
 .nvm
 /bots/node_modules/
 package-lock.json
+yarn.lock
 
 # OS X
 .DS_Store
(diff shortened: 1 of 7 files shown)


abelosorio (approved):
Looks good!

nahudalla (approved):
Excellent work! :muscle:

I didn't know `restProps` was a known pattern, but I use it all the time :stuck_out_tongue:

PR #29, opened 2019-12-12
[JB][ADDED] Added login screen to admin page and scripts to set user claims

jmbidev:
Considerations:

- To run scrips we need to add .env file in scripts folder with these keys:
  `DATABASE_URL=<databaseUrl>`
  `SERVICE_ACCOUNT_PATH=<pathToServiceAccountJsonFile>`

- From scripts folder write:
  `node setUserAsAdmin.js <email>`
  `node removeUserAsAdmin.js <email>`

image


admin-src/index.js
```diff
@@ -1,111 +1,37 @@
...
+        <Row>
+          <Button onClick={logOut}>
+            Log out
+          </Button>
+        </Row>
+        <DiagnoseResponse />
+      </>
+    )
+  } else {
+    return (<LoginScreen onUserLogin={onUserLogin} />)
```
MPucheta: There is a convention that when returning one line JSX elements, you do not use parentheses.
Parentheses are used on multiline elements.

[https://github.com/airbnb/javascript/tree/master/react#parentheses](https://github.com/airbnb/javascript/tree/master/react#parentheses)

I personally do not follow all of those conventions but it's just to mention it.

admin-src/utils/index.js
```diff
@@ -20,3 +20,18 @@ export const getDownloadURLFromImages = async (imageReferences) => (
...
+  try {
+    const user = await firebase.auth().currentUser
+
+    if (!user) return false
+
+    const result = await user.getIdTokenResult()
+
+    return Boolean(result.claims.isAdmin)
+  } catch (error) {
+    console.error(error.message)
```
MPucheta: I do not think this will work as expected. I'm not quite sure because this is the web environment, but in mobile a `console.error` will stop the execution of the app, so the `return false` will never be reached.

image


It probably never happened because `getIdTokenResult` did not fail. I suggest faking an error throw to see if this works as expected. I ping Nahuel for web specific insight

@nahudalla

MPucheta (approved):
Looks fine to me. I'd suggest that you upload the .env of scripts as part of the dev configurations in the shared folder so anyone can download it and try the scripts (We can manually set the variables, but I can see someone wondering how to set `SERVICE_ACCOUNT_PATH<pathToServiceAccountJsonFile>`).

Is it the one in `android/`? is the path like `../../../` etc

admin-src/utils/index.js
```diff
@@ -20,3 +20,18 @@ export const getDownloadURLFromImages = async (imageReferences) => (
...
+    return false
```
MPucheta: `console.error` and the like of them are commonly used to help us in development to get additional information to debug. When an error occurs, we need to solve it or provide some usable alternative to the consumer of the code.

In my eyes, this console.error is not necessary, as you correctly solve the situation.
This is nitpicky but it would be good if we decide on these things for consistency. For example, if we see a console.error we are sure that something was not taken care of properly as we require that debug information. We can always set an error to show to the user or just set an error state.

@nahudalla

admin-src/index.js
nahudalla: We don't use AirBnb's convention, we use StandardJS: https://standardjs.com/

If StandardJS doesn't say anything about the matter, our company-wide convention is to not use parenthesis in this case.

admin-src/DiagnoseResponse/index.js
```diff
@@ -0,0 +1,105 @@
...
+import { DiagnoseInfo } from '../DiagnoseInfo'
+import { getDownloadURLFromImages } from '../utils'
+import '../stylesheets/admin.css'
+
+export const DiagnoseResponse = () => {
+  const [currentDiagnose, setCurrentDiagnose] = useState(null)
+  const [diagnoses, setDiagnoses] = useState([])
+
+  useEffect(() => {
+    const attachQueryListenerForUnansweredDiagnoses = () => {
```
nahudalla, line 14: Why this function if it's called immediately?

admin-src/DiagnoseResponse/index.js
```diff
@@ -0,0 +1,105 @@
...
+  }, [])
+
+  useEffect(() => {
+    const clearScreenIfCurrentDiagnoseWasAnsweredByAnotherOne = () => {
+      if (currentDiagnose) {
+        const queryForCurrentDiagnoseInDiagnoses = diagnoses.filter(diagnose => diagnose.id === currentDiagnose.id)
+        const currentDiagnoseDoesNotExist = queryForCurrentDiagnoseInDiagnoses.length === 0
+        if (currentDiagnoseDoesNotExist) {
+          setCurrentDiagnose(null)
+        }
```
nahudalla, line 60: This can be simplified:
```javascript
const currentDiagnoseExists = diagnoses.some(({ id }) => id === currentDiagnose.id)
if (!currentDiagnoseExists) setCurrentDiagnose(null) 
```

admin-src/DiagnoseResponse/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+          setDiagnoses([...unanswered])
+        })
+    }
+
+    attachQueryListenerForUnansweredDiagnoses()
+  }, [])
+
+  useEffect(() => {
+    const clearScreenIfCurrentDiagnoseWasAnsweredByAnotherOne = () => {
```
nahudalla, line 54: Again, why this function if it's called immediately?

admin-src/utils/index.js
```diff
@@ -20,3 +20,18 @@ export const getDownloadURLFromImages = async (imageReferences) => (
...
+
+export const userHasAccess = async () => {
+  try {
+    const user = await firebase.auth().currentUser
+
+    if (!user) return false
+
+    const result = await user.getIdTokenResult()
+
+    return Boolean(result.claims.isAdmin)
```
nahudalla: You can directly return `result.claims.isAdmin` here.

admin-src/utils/index.js
nahudalla: I agree, no need to log the error if you handled it (not allowing access).

scripts/removeUserAsAdmin.js
```diff
@@ -0,0 +1,29 @@
...
+    console.error(error.message)
+  }
+  process.exit(0)
+}
+
+admin.auth().getUserByEmail(mail).then(
+  user => setUserAsAdmin(user)
+).catch(
+  error => console.error(error.message)
+)
```
nahudalla: Here I'd suggest writing this as follows, for readability:
```javascript
admin.auth().getUserByEmail(mail)
  .then(setUserAsAdmin)
  .catch(console.error.bind(console))
```

scripts/removeUserAsAdmin.js
```diff
@@ -0,0 +1,29 @@
...
+
+const setUserAsAdmin = async (user) => {
+  try {
+    await admin.auth().setCustomUserClaims(user.uid, { isAdmin: false })
+    console.log('User admin privileges succesfully removed')
+  } catch (error) {
+    console.error('Could not remove admin privileges for user')
+    console.error(error.message)
+  }
+  process.exit(0)
```
nahudalla, line 22: Is this necessary? Isn't the default exit code `0`?

scripts/setUserAsAdmin.js
```diff
@@ -0,0 +1,29 @@
...
+    console.error('Could not add user as admin')
+    console.error(error.message)
+  }
+  process.exit(0)
+}
+
+admin.auth().getUserByEmail(mail).then(
+  user => setUserAsAdmin(user)
+).catch(
+  error => console.error(error.message)
```
nahudalla: Same comments as `removeUserAsAdmin`

scripts/removeUserAsAdmin.js
```diff
@@ -0,0 +1,29 @@
...
+  process.exit(1)
+}
+
+const mail = process.argv[2]
+
+initFirebaseAdmin()
+
+const setUserAsAdmin = async (user) => {
+  try {
+    await admin.auth().setCustomUserClaims(user.uid, { isAdmin: false })
```
nahudalla, line 16: I'd remove the property `isAdmin` completely if the user is not an admin:
```javascript
await admin.auth().setCustomUserClaims(user.uid, { })
```

nahudalla (changes requested):

nahudalla (approved):

PR #30, opened 2019-12-13
MP/ Adapt signup login to v4 wireframe

MPucheta:
- Changed layout and flow

- Removed login screen (it was not necessary anymore, as social networks buttons are in sign up and login)

- Added privacy policy text in sign up

- Abstracted login process to a single login function that uses Login methods with a take on Dependency Injection as suggested from @jmbidev which seemed like a good idea. SignUp code was left as it is though (it's different semantically, abstracting it more would add indirection)

- UsernameRequest now **ASSUMES** that a user is logged in, so logging in and writing new user information are two separated tasks. If you end up in UsernameRequest, it's because you are a new user.

- Signup and login uses the same social network button, there is practically no gain from separating both processes. We could put a check if the user already has an account with that email and display a text, but it's just putting rocks in the way because the user clearly wants to login, the natural flow would be going to the login and pressing the same button.

It's just for completeness, but here is the signup which mimics the wireframe.

image


src/authenticationService/index.js
```diff
@@ -11,47 +11,46 @@ export const getCurrentUserUID = () => {
...
 
-    return firebaseUserCredential
+export const createUserWithEmailAndPassword = async (email, password) => {
+  try {
+    const firebaseCredential = await firebase.auth().createUserWithEmailAndPassword(email, password)
+    await MessagingService.enableNotificationsForUser()
+    return firebaseCredential
   } catch (error) {
     throw new AuthenticationError(error.message)
   }
```
nahudalla: - This can be extracted into another `authenticationMethod` (as you did for `GoogleSignInMethod` and `EmailSignInMethod`) called `EmailSignUpMethod`.
- I think the names of the methods are pretty explanatory without the `Method` suffix: `GoogleSignIn`, `EmailSignIn` and `EmailSignUp`.
- If you extract this to an authentication method, the `login` method of the method classes can be changed to `authenticate` to be more generic.

What do you think?

src/authenticationService/signInMethods/index.js
```diff
@@ -0,0 +1,36 @@
+import * as firebase from 'firebase'
+import { GoogleSignin } from 'react-native-google-signin'
+import googleConfig from '~/configs/google'
+import AuthenticationError from '~/AppErrors/AuthenticationError'
+
+export class GoogleSignInMethod {
```
nahudalla: I would extract each of the authentication methods to its own file.

src/screens/authentication/SocialNetworks/GoogleButton/index.js
```diff
@@ -1,14 +1,41 @@
...
+
+      setSigning(false)
+      if (username) {
+        NavigationService.navigate('MainApp')
+      } else {
+        NavigationService.navigate('UsernameRequest')
+      }
+    } catch (error) {
+      setSigning(false)
+    }
```
nahudalla: Avoid repeating `setSigning(false)` by calling it in the `finally` block:
```javascript
try {
  // something that may fail
} catch (err) {
  // do something with the error
} finally {
  // do something whether an error was thrown or not
}
```
Just as you can omit the `finally` block, you can also omit the `catch` block. The following is valid:
```javascript
try {
  // something that may fail
} finally {
  // do something whether an error was thrown or not
}
```

nahudalla (changes requested):

src/screens/authentication/SocialNetworks/GoogleButton/index.js
MPucheta: Repetition was mainly because NavigationService prompts the component to be unmounted and a warning appears showing that "setting state in an unmounted component is a no-op". So it was necessary to set the state before navigating. I can diminish repetition with a boolean workaround though.

src/authenticationService/index.js
MPucheta: Those are great suggestions. I considered it necessary to separate between `signup` and `login` as they have different semantics, but it's true, both can be extracted into `authenticate` and remain clean.

src/screens/authentication/SocialNetworks/GoogleButton/index.js
nahudalla: The solution here, in my opinion, would be extracting the variable and the navigation outside of the `try/catch`:
```javascript
setSigning(true)

let username
try {
  const firebaseCredential = await AuthenticationService.googleLogin()
  username = await DatabaseService.queryUsernameFromEmail(firebaseCredential.user.email)
} finally {
  setSigning(false)
}

if (username) {
  NavigationService.navigate('MainApp')
} else {
  NavigationService.navigate('UsernameRequest')
}
```

src/authenticationService/index.js
```diff
@@ -11,47 +11,44 @@ export const getCurrentUserUID = () => {
...
-    const googleSignInData = await GoogleSignin.signIn()
+    return firebase.auth().currentUser
+  } catch (error) {
+    throw new AuthenticationError('Not logged in.')
+  }
+}
 
-    return googleSignInData
+const authenticate = async (AuthenticationMethod) => {
+  // SigninMethod should be either EmailLogin, GoogleLogin, or any other that we may need.
```
nahudalla: Please remove this comment. This should be obvious when reading the code that you need to pass an object with an `authenticate` method (Duck Typing). (Also, it's already outdated. It's referring to a variable that changed its name. That's one of the reasons why comments are bad)

nahudalla (changes requested):

nahudalla (approved):

PR #33, opened 2019-12-19
[MP]Adding-images-visualization-on-request

MPucheta:
- Added Image review as close as possible to the wireframes

- This is a difficult screen to refactor, ideas are to do it are very much appreciated. It's also a heavy performance hitter (the navigation header is static to the component, so a workaround must be done to redefine functions and bind them to the component state). I think a more direct way could be done using Class-based components. Regardless of this, this task inherently requires a lot of state bindings.

- This implementation assumes some data from the screen Jeremias is doing, as we are doing this in parallel, I decided to use a "mock" on images to do the implementation.

- Screencast is added.

preview


src/screens/MainApp/DiagnoseRequest/ImageReview/ConfirmButton/index.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import { Text, Button } from '~/components'
+
+const Error = ({ errorState }) => (
+  errorState &&
+    <Text
+      colorVariant='secondary'
+      style={{ alignSelf: 'center' }}
+    >Selecciona al menos 3 imagenes para continuar!
```
abelosorio: I don't personally like this syntax very much:
```javascript
<Component
  prop1={value1}
  prop2={value2}
>Content
</Component>
```
I think it's clearer this way:
```javascript
<Component
  prop1={value1}
  prop2={value2}
>
  Content
</Component>
```

(Just an opinion)

src/screens/MainApp/DiagnoseRequest/ImageReview/index.js
```diff
@@ -0,0 +1,136 @@
...
+  const getMorePhotos = () => {
+    const filterDuplicates = newImages.filter(element => !images.includes(element))
+    setImages([...images, filterDuplicates[0]])
+  }
+
+  const FlatListItem = ({ item }) => {
+    const itemIndex = images.indexOf(item)
+    const lastIndex = images.length - 1
+    const selected = images[activeIndex] === item
+    if (itemIndex === lastIndex && images.length < MAX_IMAGES) {
```
abelosorio: Let's leave a blank line before blocks like this and always before `return` statements (the only exception is when the `return` statement is the only statement within a block).

src/screens/MainApp/DiagnoseRequest/ImageReview/index.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  useEffect(() => {
+    if (images.length < MIN_IMAGES) {
+      setSubmitError(true)
+    } else {
+      setSubmitError(false)
+    }
+
+    const deleteImage = () => {
```
abelosorio: I think this function could be more readable if you extract the content of the conditions out to separate functions. If I read correctly (I did it quickly) the first condition deletes a single image and the second one multiple images, am I right? If so, it could be much clearer this way:

```javascript
if (images.length === 1) {
  deleteSingleImage(images)
} else {
  deleteMultipleImages(images)
}
```

Related but to exactly for this case, just material to think about: https://www.youtube.com/watch?v=os6ssw1qylU

src/screens/MainApp/DiagnoseRequest/ImageReview/index.js
```diff
@@ -0,0 +1,136 @@
...
+      />
+      <ConfirmButton errorState={submitError} onConfirm={confirmRequest} />
+    </View>
+  )
+}
+
+ImageVisualization.navigationOptions = ({ navigation }) => ({
+  headerRight: () => (
+    <>
+      <Button variant='alpha' onPress={navigation.getParam('changeImage')}><Icon name='redo' size={18} color='black' /></Button>
```
abelosorio: You're closing the component twice, is that for a special reason?

abelosorio (changes requested):

src/screens/MainApp/DiagnoseRequest/ImageReview/index.js
MPucheta: Both paths delete a single image. The division was necessary to Navigate to another screen when no images are present (the user deleted all of them) and because 

`images.filter((val, index) => index !== activeIndex)` 

would fail when `images.length === 1` as the current selected index is the only one.

I had problems implementing this though, is there a convention when modifying an array that is part of the state?

I wanted to use `images.splice(activeIndex, 1)` but it was just modifying the state directly. I also thought about doing a copy and then splice on the copy, but it seemed overkill.

Regardless of this, I think I can extract functions like that.

src/screens/MainApp/DiagnoseRequest/ImageReview/index.js
MPucheta: I will write this different. There is actually an Icon component which is closed before the Button tag.

nahudalla (changes requested):
Refactor as discussed in our call today.

nahudalla (approved):

jmbidev (approved):
Seems nice to me. 
One thing to consider is the using the flex property instead of widthPercentageToDP or heightPercentageToDP, also depending on the layout design you can set width and height with percentage values.

src/screens/MainApp/ImageReview/index.js
```diff
@@ -0,0 +1,98 @@
...
+  const deleteActiveIndexImage = () => {
+    const wantedImages = images.slice()
+    wantedImages.splice(activeIndex, 1)
+    setImages(wantedImages)
+
+    return wantedImages
+  }
+
+  const setSubmitErrorIfImageCountBelowMinimum = () => {
+    if (images.length < MIN_IMAGES) {
```
abelosorio, line 33: This could be clearer if you do:
```javascript
return setSubmitError(images.length < MIN_IMAGES)
```

src/screens/MainApp/NoPhotoDisclaimer/index.js
```diff
@@ -0,0 +1,30 @@
...
+
+const goToCallToAction = () => {
+  console.log('Going to Call To Action')
+}
+
+const NoPhotoDisclaimer = () => (
+  <View style={styles.container}>
+    <Image source={DrCannabis} style={styles.image} />
+    <Text colorVariant='gray'>
+      Ya no tenés fotos para enviar
```
abelosorio, line 15: Not sure how React handles accents. Shouldn't we be using html entities instead?

cc @nahudalla

abelosorio (approved):
A few non-blocking comments.

src/screens/MainApp/NoPhotoDisclaimer/index.js
nahudalla: I think it should handle any valid JavaScript string here, since this is transpiled into a string by Babel. JavaScript strings are UTF-16, so it should work correctly in my opinion.

src/screens/MainApp/ImageReview/index.js
nahudalla: @MPucheta please make this quick change and let me know so that I can merge the PR.

PR #34, opened 2019-12-21
[MP] MyDiagnoses-Screen

MPucheta:
- Implemented 'mis solicitudes' screen as it shows in the wireframe

- I added a cut to the text (dots at the end), as too much description would result in the cards beings too big and contradicting what the wireframe intends to show.

- I did not include the 'Abierto' and such indications. It seems more like the responsibility of the 'diagnose details' screen (I think we would need to use the cache too).

Screencasts:

image
solicitudes


src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -0,0 +1,81 @@
...
+  const [diagnoses, setDiagnoses] = useState(null)
+  const [downloadingDiagnoses, setDownloadingDiagnoses] = useState(null)
+  const [refetchTrigger, toggleRefetch] = useState(false)
+  const [showNoDiagnosesScreen, setShowNoDiagnosesScreen] = useState(false)
+
+  const refetchDiagnoses = () => {
+    // Tab navigator doesn't unmount components, we need this to refetch need data when pressing this tab.
+    // alternative is to add a query snapshot listener and a MobX store
+    toggleRefetch(!refetchTrigger)
+  }
```
MPucheta: I really do not like this. I believe it's ok for now and it really doesn't incur in more database hits (there is a window of 30 minutes of free hits if the data did not change, I could not find this in the docs though, so I ping @nahudalla for insight) But it seems hacky to force a re-render this way. I think the best way to do this is with mobx and introducing reactiveness, but it's a little expensive on time.

nahudalla: Maybe this will solve the problem: https://reactnavigation.org/docs/en/next/use-focus-effect.html

_We're using React Navigation, right?_

MPucheta: I will give it a shot

src/mixins/diagnose/index.js
```diff
@@ -0,0 +1,7 @@
+export const sortDiagnosesByMostRecentCreation = (diagnoses) => (
+  diagnoses.sort((a, b) => b.createdAt.toMillis() - a.createdAt.toMillis())
+)
+
+export const sortDiagnosesByMostRecentUpdate = (diagnoses) => (
+  diagnoses.sort((a, b) => b.updatedAt.toMillis() - a.updatedAt.toMillis())
+)
```
nahudalla: What do you think about generalizing the sorting this way?

```javascript
const sortAscending = (array, getSortValue = a => a) => (
  array.sort((a, b) => getSortValue(a) - getSortValue(b))
)

export const sortDiagnosesByMostRecentCreation = (diagnoses) => (
  sortAscending(diagnoses, a => a.createdAt.toMillis())
)

export const sortDiagnosesByMostRecentUpdate = (diagnoses) => (
  sortAscending(diagnoses, a => a.updatedAt.toMillis())
)
```

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -0,0 +1,81 @@
...
+    try {
+      const downloadedData = await DatabaseService.getDiagnosesFromCurrentUser()
+      const sortedData = sortDiagnosesByMostRecentCreation(downloadedData)
+      const renderedDiagnoses = await renderDiagnoses(sortedData)
+
+      setDiagnoses(renderedDiagnoses)
+    } catch (error) {
+      setDiagnoses(null)
+    }
+    setDownloadingDiagnoses(false)
```
nahudalla: Please move this line to the `finally` block

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -0,0 +1,81 @@
...
+    if (downloadingDiagnoses === false) {
+      if (noDiagnoses()) {
+        setShowNoDiagnosesScreen(true)
+      }
+    }
+  }, [downloadingDiagnoses])
+
+  if (showNoDiagnosesScreen) {
+    return <NoDiagnoses />
+  }
```
nahudalla: This doesn't need an effect nor a state:
```javascript
import { isEmpty } from 'lodash'

// ...

if (downloadingDiagnoses) return null // or a loading indicator?
if (isEmpty(diagnoses)) return <NoDiagnoses />

// ...
```

src/screens/MainApp/MyDiagnoses/renderUtilities/index.js
```diff
@@ -0,0 +1,43 @@
...
+    <Diagnose
+      key={key}
+      thumbnail={thumbnail}
+      firebaseTimestamp={diagnose.createdAt}
+      description={diagnose.text}
+      answerQuantity={answerQuantity}
+    />
+  )
+}
+
```
nahudalla: I don't like this asynchronous rendering. Rendering is meant to be a synchronous process. In the future, React will introduce concurrent rendering. I don't know how well asynchronous rendering will play with concurrent rendering. If you need to perform asynchronous actions as part of the rendering process of a component, you have to do so when React tells you. That's the purpose of `useEffect`.

nahudalla (changes requested):

src/mixins/diagnose/index.js
MPucheta: It's quite good. There may be some cases that unexpected behavior occur if the `getSortValue` does not return something that makes sense with the `-` operator, but it's up to the developer to take care of this.

I will implement this right away

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -0,0 +1,85 @@
...
+  }, [refetchTrigger])
+
+  if (isEmpty(diagnoses)) {
+    return <NoDiagnoses />
+  }
+
+  return (
+    <View style={styles.container}>
+      <ForceRerenderOnNavigation resetStateFunction={refetchDiagnoses} />
+      {downloadingDiagnoses && <ActivityIndicator size='large' />}
```
MPucheta, line 64: I left this indicator here as returning the whole indicator outside would cause a white flicker in the screen (there is a temporal frame between forcing the re-render and the diagnose fetching, so the user could still see the diagnoses for 1 second before having them disappear and appear again, with this here that does not happen and the user gets a better experience.... hopefully)

nahudalla (approved):

jmbidev (approved):

PR #36, opened 2019-12-24
MP/ Add detailed diagnose screen

MPucheta:
**I put this as a draft because MyDiagnoses is not merged yet**

- Added detailed diagnose screen

- Added some more mixins that turned out to be repeated throughout the code

- These changes take into considerations the possibility of more answers and more than one image.

- Added pluralize

**UPDATED SCREENCAST**
preview

**SUPPORT FOR MORE THAN ONE IMAGE**

image


src/components/Text/index.js
```diff
@@ -51,7 +52,7 @@ const styles = StyleSheet.create({
 Text.propTypes = {
   style: PropTypes.object,
   colorVariant: PropTypes.oneOf(['primary', 'secondary', 'black', 'white', 'gray', 'gray2', 'gray3']),
-  fontVariant: PropTypes.oneOf(['h1', 'h2', 'h3', 'description', 'body', 'date'])
+  fontVariant: PropTypes.oneOf(['h1', 'h2', 'h3', 'title', 'description', 'body', 'date'])
 }
```
nahudalla: Be careful, this component is growing too much. Maybe what we need here are different components that _extend_ `Text` (by composition). Each new component would only apply its specific styles.

src/mixins/date/index.js
```diff
@@ -0,0 +1,5 @@
+import moment from 'moment'
+
+export const firebaseTimestampToDate = (timestamp) => (
+  moment(timestamp.seconds * 1000)
+)
```
nahudalla: This returns a `Moment` object, not a `Date` object. Consider making that clear on the function name.

src/mixins/text/index.js
```diff
@@ -0,0 +1 @@
+export const addPluralizationIf = (condition) => condition ? 's' : ''
```
nahudalla: Is there any library that does this automatically? I know of this one, but I don't know if it works for spanish: ://github.com/blakeembrey/pluralize

src/screens/MainApp/DetailedDiagnose/Carousel/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+      await Promise.all(
+        images.map(async (imageUID, index) => {
+          const url = await getURL(imageUID)
+          urls.push(url)
+        })
+      )
+
+      setImageExternalURLs(urls)
+      setFetchingURLs(false)
```
nahudalla: If something fails, this statement will not execute and the state won't change to `fetchingURLs=false`. Consider wrapping the `await Promise.all` in a `try/catch/finally` block, and put this inside `finally`.

src/screens/MainApp/DetailedDiagnose/index.js
```diff
@@ -0,0 +1,47 @@
...
+  const date = firebaseTimestampToDate(diagnose.createdAt)
+  const answers = []
+
+  if (diagnose.answer) {
+    answers.push(
+      {
+        answer: diagnose.answer,
+        by: diagnose.answeredBy
+      }
+    )
```
nahudalla: Please place the curly brackets in the same line of the parenthesis:
```javascript
answers.push({
  answer: diagnose.answer,
  by: diagnose.answeredBy
})
```

src/screens/MainApp/DetailedDiagnose/index.js
```diff
@@ -0,0 +1,47 @@
...
+        <Datetime date={date} />
+        <Description description={diagnose.text} />
+        <View>
+          <Text fontVariant='title' colorVariant='black'>
+            {answers.length} Respuesta{addPluralizationIf(answers.length !== 1)}
+          </Text>
+          <Text />
+        </View>
+        <AnswerList data={answers} />
+        {answers.length === 0 && <NoAnswersYet />}
```
nahudalla: Use `isEmpty` from `lodash` here.

nahudalla (commented):
Please let me know what you think about these comments.

src/screens/MainApp/DetailedDiagnose/Carousel/index.js
MPucheta: `getURL` returns the real URL, but if it fails, returns null (the carousel then shows a placeholder image).

I do not think there is something else that could fail but I'l will add the `try/catch` for safety

src/mixins/text/index.js
MPucheta: I can give it a try. There is a pluralize-es but it has not been updated for 4 years now....

abelosorio: FYI I've been using this library https://www.npmjs.com/package/pluralize and it's been good so far.

MPucheta: It worked for 'respuesta', I suspect there could be some corner cases but all good for now.

nahudalla (approved):

PR #39, opened 2019-12-30
Jb/diagnose request screens connection

jmbidev:
Nothing much to say here. It should work as in wireframes except gallery screen.


MPucheta (approved):

src/imageService/index.js
```diff
@@ -1,39 +1,42 @@
...
+    const reducer = (a, c) => {
+      a.push(c.path)
+      return a
+    }
+
+    if (Array.isArray(images)) {
+      return images.reduce(reducer, [])
+    } else {
+      return [images].reduce(reducer, [])
+    }
```
nahudalla: You're violating DRY here, and also you want to `map` the values, not `reduce`:

```javascript
if (!Array.isArray(images)) images = [images]

return images.map(img => img.path)
```

src/screens/MainApp/DiagnoseRequest/navigators/MainNavigator.js
```diff
@@ -0,0 +1,35 @@
...
+
+MainNavigator.navigationOptions = ({ navigation }) => {
+  let tabBarVisible = true
+
+  if (navigation.state.index > 0) {
+    tabBarVisible = false
+  }
+
+  return { tabBarVisible }
+}
```
nahudalla: 🤔 why is the tab bar hidden if the `navigation.state.index` is `> 0`? This code is not clear enough. Try using a function/variable for the conditions, so that you can use a descriptive name: (the following name is only an example)
```javascript
let tabBarVisible = true
const isFirstScreen = navigation.state.index === 0

if (!isFirstScreen) tabBarVisible = false

return { tabBarVisible }
```
When I read the code above I interpret `The tab bar is visible on all screens, except the first one`. I don't need to know the specifics of the condition, but I can find it quickly if I want.

src/screens/MainApp/DiagnoseRequest/screens/ImageReview/index.js
```diff
@@ -1,28 +1,45 @@
...
   const [activeIndex, setActiveIndex] = useState(0)
   const [submitError, setSubmitError] = useState(false)
-  const [images, setImages] = useState(requestImages.slice())
+  const [images, setImages] = useState(props.navigation.state.params.images)
+  const [showImageSelection, setShowImageSelection] = useState(false)
 
-  const confirmRequest = () => {}
+  useEffect(() => {
+    setImages(props.navigation.state.params.images)
+  }, [props.navigation.state.params.images])
```
nahudalla: You're using this chain of accessors three times. I need to change three places if I want to change the way I receive the images. Consider using a variable for this:
```javascript
const imagesFromPreviousStep = props.navigation.state.params.images

// use imagesFromPreviousStep
```

src/screens/MainApp/DiagnoseRequest/screens/ImageReview/index.js
```diff
@@ -1,28 +1,45 @@
...
+
+  const confirmRequest = () => {
+    NavigationService.navigate('DescriptionRequest', { images: images })
```
nahudalla: No need to repeat the `images` variable: `{ images: images }` is equivalent to `{ images }`

src/screens/MainApp/DiagnoseRequest/screens/ImageReview/index.js
```diff
@@ -44,18 +61,23 @@ const ImageVisualization = (props) => {
...
       }
     }
 
-    const changeImage = () => {
-      const changedImageArray = images.slice()
-      changedImageArray[activeIndex] = newImages[activeIndex]
-      setImages(changedImageArray)
+    const changeImage = async () => {
+      try {
+        const newImage = (await ImageService.openCamera())[0]
```
nahudalla: 🤔 this looks ugly. What if we use array destructuring to extract the first element?
```javascript
const [ newImage ] = await ImageService.openCamera()
```

src/screens/MainApp/DiagnoseRequest/screens/MainScreen/index.js
```diff
@@ -7,19 +7,16 @@ export const MainScreen = ({ navigation }) => {
   const [showImageSelection, setShowImageSelection] = useState(false)
 
   const onImagesSelected = (images) => {
-    const reduced = images.reduce((a, c) => {
-      a.push(c.path)
-      return a
-    }, [])
     setShowImageSelection(false)
+    navigation.navigate('ImageReview', { images: images })
```
nahudalla, line 11: Same here: `{ images: images }` is equivalent to `{ images }`

src/screens/MainApp/DiagnoseRequest/screens/NoPhotoDisclaimer/index.js
```diff
@@ -0,0 +1,44 @@
...
+import { Text, Button, ImageSelection } from '~/components'
+import DrCannabis from '~/assets/images/DrCannabis.png'
+import styles from './styles'
+
+const NoPhotoDisclaimer = ({ navigation }) => {
+  const [showImageSelection, setShowImageSelection] = useState(false)
+
+  const onImagesSelected = (images) => {
+    setShowImageSelection(false)
+    navigation.navigate('ImageReview', { images: images })
```
nahudalla: Same here: `{ images: images }` is equivalent to `{ images }`

nahudalla (changes requested):
Almost there! Only a few cosmetic code changes.

nahudalla (approved):

PR #40, opened 2019-12-30
MP/ Create new reusable components

MPucheta:
- These adhere better to the open-closed and single responsibility principles.

- We decided to stick to a 'variant' approach in the text color property, as doing another component or higher order component wouldn't provide benefits and would in turn make the code too verbose.
(we have `<Description primary />` vs having some kind of injector component just to add color). It makes more sense to do this separation in the buttons, as I can imagine doing specific things to each one (for example `box shadow`), but I did not see a practical gain from complicating Text-based components.

- Added the buttons that were present in the guides, specific Buttons (like the ProgressButton too) should be added when they are needed (although it's just a copy-paste away, it adheres to YAGNI)

- Previous Text/Button components were deleted, so styling in different screens is a mess (on purpose, because we have to change them with the new designs)


jmbidev (approved):

src/components/index.js
```diff
@@ -1,4 +1,22 @@
...
+} from './texts'
+
+export {
+  Button,
+  GrayButton,
+  PrimaryButton,
+  AlphaButton,
+  BlackButton,
+  ImageButton
+} from './buttons'
```
nahudalla: 🤔 I don't like these `index.js` files that only export other things...
- First, you could've written this as `export * from './texts` (same for buttons).
- Second, these index files are another thing to keep maintained. It's easier to just import the thing you need from the appropriate path: `import GrayButton from '~/components/buttons/GrayButton'`

src/components/texts/Date.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { StyleSheet } from 'react-native'
+import Text from './Text'
+import { theme } from '~/constants'
+
+const Date = (props) => {
```
nahudalla, line 6: Are all dates rendered with the same style on the app? Why not just as text? I don't know, this component doesn't seem as it's really needed.

nahudalla (changes requested):

src/components/texts/Date.js
MPucheta: Yes. Dates have a specific size and a specific font. While normal text uses `Red Hat Regular`, dates use `Red Hat Medium` and with a lower size.

nahudalla (approved):

PR #41, opened 2019-12-31
MP/ Adapt theme file to styleguide

MPucheta:
- This includes fonts, sizes and colors

- 16k lines are from package-lock.json


jmbidev (approved):

react-native.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  project: {
+    ios: {},
+    android: {}
+  },
+  assets: ['./src/assets/fonts/']
```
nahudalla, line 6: Do we really need to ship all these fonts with the APK? Why so many? Can't we download them dynamically if they're hosted on Google Fonts? Can't we use some package like [react-native-vector-icons](https://github.com/oblador/react-native-vector-icons) so that we only ship the icons we use?

I'm worried about the final size of the APK.

MPucheta: The huge number of fonts were already bundled into the app, they appear now because I had to use `react-native link` to add `Red Hat Display` and `Raleway`. We already use `react-native-vector-icons`, in particular, `FontAwesome5`. 

I researched on online loading, there were some libraries but I do not think they are battle tested (also last updates were 6 months - 1 year old)

https://github.com/cawfree/react-native-custom-fonts
https://github.com/eVisit/react-native-dynamic-fonts

I, particularly, do not think this is a problem as the base `.ttf` are only a couple of kB.

nahudalla (approved):

nahudalla:
@MPucheta please solve the conflicts

PR #42, opened 2020-01-03
MP/Adapt signup and login to styleguide

MPucheta:
- Extracted common functionality into Auth component, which has a specific layout and components. Form component inside Auth ended up a little big, while I can extract some things here and there, I prefer not to to avoid multiple levels of nesting.

- Added SVG support for the google button and improved it's error handling capabilities.

- Introduced some changes to theme due to shadows and scaling (already talked with Jeremias)

I attach signup, which is practically the same as login

image


src/screens/authentication/components/Auth/Form/index.js
```diff
@@ -0,0 +1,129 @@
...
+
+const Form = ({
+  initialValues,
+  handleSubmit,
+  schema,
+  error,
+  authenticating,
+  credentialText,
+  submitText
+}) => {
```
nahudalla: If you created this form to encapsulate the fields of the authentication form, then `initialValues` and `schema` are internal to the components. Otherwise, when I use this component I need to know what fields it renders to build the correct `initialValues` / `schema`.

src/screens/authentication/components/Auth/Separator/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import Separator from '~/components/Separator'
+import { verticalScale } from 'react-native-size-matters'
+
+const AuthSeparator = ({ children }) => (
+  <Separator style={{
+    position: 'absolute',
+    top: verticalScale(457)
+  }}
+  >
```
nahudalla, line 10: Is this the correct indentation? What does Standard say about it?

src/screens/authentication/utils/index.js
```diff
@@ -1,12 +1 @@
...
 export const isValidEmail = (text) => text.includes('@')
-
-export const enableNotificationsForUser = async () => {
-  try{
-    await MessagingService.checkForPermissions()
-    await MessagingService.saveFCMTokenForCurrentUser()
-  } catch (error) {
-    console.error(error.message)
-  }
-}
```
nahudalla, line 12: Was this old code that is not used anymore? Where are notifications enabled now?

nahudalla (changes requested):

src/screens/authentication/utils/index.js
MPucheta: Yes, we abstracted it into the authentication process (emailSignUp, emailLogin and googleLogin)

src/screens/authentication/components/Auth/Separator/index.js
MPucheta: It did not mark it as an error, putting the `>` actually was marked as an error

image

src/screens/authentication/components/Auth/Form/index.js
MPucheta: While this is true, `schema` is only used in the case of `SignUp`, as it would disturb the user in the case of the Login (text is displayed in case of missing fields), there is also some references to values in the handling method. I can move the `initialValues` and `schema` inside the component and provide a prop `doNotValidate`, but I think the artificial coupling will continue to exist due to code and style reuse.

nahudalla (approved):

PR #44, opened 2020-01-06
MP/ Adapt username request to styleguide

MPucheta:


src/screens/authentication/UsernameRequest/index.js
```diff
@@ -1,22 +1,61 @@
...
+  } else {
+    return (
+      <PrimaryButton
+        onPress={onPress}
+      >
+        <Description white>Continuar</Description>
+      </PrimaryButton>
+    )
+  }
+}
```
nahudalla: The `value` prop is only used for the `if` statement?

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -62,31 +97,31 @@ const UsernameRequest = (props) => {
...
-          source={DrCannabis}
-        />
-        <Text fontVariant='description' style={styles.information}>Antes de iniciar sesión vamos a necesitar un nombre para tu usuario.</Text>
-        <Text fontVariant='h1' style={styles.usernameTag}>Usuario</Text>
+        <Subtitle style={styles.title}>Creá tu nombre de usuario</Subtitle>
+        <Description
+          gray
+          style={styles.description}
+        >
+        Escribí un nombre de usuario para identificarte dentro de la plataforma
```
nahudalla: Wrong indentation here.

nahudalla (changes requested):

src/screens/authentication/UsernameRequest/index.js
MPucheta: Yes

src/screens/authentication/UsernameRequest/index.js
MPucheta: Standard did not mark it. Should I put the text two spaces to the right?

nahudalla: Yes, I like increasing indentation inside every tag. I don't know if standard has a rule for that.

src/screens/authentication/UsernameRequest/index.js
nahudalla: :thinking: I think that prop should be a `boolean` with a more descriptive name, like `disabled`.

nahudalla (approved):

PR #49, opened 2020-01-07
MP/ Adapt image review to styleguide

MPucheta:
**This is branched from the Carousel branch, I put up a draft until merged, although some changes to the component were made, it retains it's interface, so in the future nothing breaks when merged**

- Adapted to styleguide

- Introduced some UX improvements

- Introduced changes to tabBarVisibility, so it shows the tab bar in a subset of the screens instead of only one (as the styleguide requires the tabBar to be visible in the `NoPhotoDisclaimer` and `FinishRequest`)

A complete screencast is attached showing:

- adding images [with the check for duplicates for gallery], 

- deleting images and 

- deleting all images to prompt NoPhotoDisclaimer

[styles for ImageSelection and MainScreen are not merged yet, so they are not styled]
preview


src/screens/MainApp/DiagnoseRequest/screens/ImageReview/components/ConfirmButton/index.js
```diff
@@ -0,0 +1,27 @@
...
+  <>
+    {errorState
+      ? (
+        <GrayButton
+          style={styles.button}
+        >
+          <Description white>Confirmar</Description>
+        </GrayButton>
+      )
+      : (
```
nahudalla: I don't know if Standard says anything about it, but I like this in one line: `) : (`, it's just the same as a `} else {`.

nahudalla (approved):

PR #50, opened 2020-01-08
MP/ Adapt MyDiagnoses to styleguide

MPucheta:
- This includes the addition of the header component that changes when scrolling the flat list. General style changes and refactorings to use the existings components were also applied

Problems that I anticipate:

- `moment` has different libraries depending of the locality. One solution I wanted to do is format as follows `date.formate('LLL')` that automatically takes care of that, but you somewhat lose the ability to know which Strings manipulations you have to do, so I left it with the usual formatting and taking Spanish language.

image


src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -40,35 +45,55 @@ const MyDiagnoses = () => {
...
-      {diagnoses &&
-        <FlatList
-          data={diagnoses}
-          showsVerticalScrollIndicator={false}
-          renderItem={({ item }) => item}
-          keyExtractor={item => String(diagnoses.indexOf(item))}
-        />}
-    </View>
+    <>
+      <View style={styles.container}>
```
nahudalla, line 62: There's no need for the fragment here (`<>`). Everything is wrapped by the `View` component.

nahudalla (approved):

PR #52, opened 2020-01-09
JB/ Add Settings/Profile screen styles as in designs

jmbidev:
Nothing much to say here, I had to change a few things in PrivacyPolicy screen and stacks to get the right flow of the screens. 
There will be code duplication when this branch and jb/description-upload-screens-designs branch get merged. I'll clean it when merging is done.  Everything should work fine.

Something to remark. I changed the text 'Salir' to 'Cerrar sessión' because it seemed more explanatory for me and it was on the other screen design. If that doesn't please you I'll change it again to 'Salir'.

Screenshot_1578600952


src/screens/MainApp/Profile/index.js
```diff
@@ -50,23 +59,28 @@ const Profile = () => {
 
   return (
     <View style={styles.container}>
-      <Text fontVariant='h1' colorVariant='black' style={styles.title}>Perfil</Text>
-      <ScrollView contentContainerStyle={styles.list}>
+      <ScrollView>
+        <Title black style={styles.title}>Mi cuenta</Title>
```
MPucheta, line 63: `cuenta` is capitalized in the design --> `Cuenta`

src/screens/MainApp/Profile/styles.js
```diff
@@ -1,46 +1,44 @@
...
   container: {
     flex: 1,
-    flexDirection: 'column'
+    flexDirection: 'column',
+    padding: theme.sizes.margin * 2
   },
   title: {
-    textAlign: 'center',
-    padding: moderateScale(10)
+    fontSize: 24,
```
MPucheta: Please consider using scaling units for `fontSizes` too or prefer the theme one. If it seems small compared to the design then we change that value in the theme.

src/screens/MainApp/Profile/index.js
jmbidev: It is intentionally or mistaken?

src/screens/MainApp/Profile/styles.js
jmbidev: I'll change it to use scaling units. The theme doesn't contain that size.

src/screens/MainApp/Profile/index.js
nahudalla: In spanish we don't capitalize every word in titles. `Mi cuenta` is OK. In English it'd be `My Account`. Maybe Agos mixed languages and didn't notice.

src/screens/MainApp/Profile/styles.js
nahudalla: @jmbidev don't all titles have to be the same size? Why don't we add this to the theme?
cc @MPucheta

nahudalla (changes requested):
Please let's finish [this discussion](https://github.com/WeDevelop-ARG/cannabis-app/pull/52#discussion_r365246841) before merging.

src/screens/MainApp/Profile/styles.js
jmbidev: I only needed this size in this screen. The other titles are defined as h1 h2 h3 y h4 on theme with sizes 20, 18, 16, 14 respectively.

@nahudalla  @MPucheta

MPucheta: Titles do have the same size. Which is the the one applied in the `Title` component using `h1` from the theme (which has a size of `moderateScale(20,1)`. In the case that value seems small compared to the designs, we can change it in only one place (although the current one seems fine by me, it's the one used in the Onboarding, Auth, MyDiagnoses, etc).

nahudalla: @jmbidev What I'm asking is _why_ does this screen need a different title size than the other screens that Mauro enumerated above. All titles have to be the same size on all screens. If they're different than the design, then we have to change all titles, not just this one.

jmbidev: Because the size used on the screen design was different from the sizes defined in theme. In the design Agos used size 24 and the defined sizes are the ones I enumerated above. Also in the styleguide that Agos created there was no size 24 defined, I think that could be a reason because it's not defined on theme. 

@nahudalla

jmbidev: There is not big trouble I can remove size property and modify h1 to be size 24

jmbidev: The problem here was that I already used Title element on those screens with font 20 titles (this happens because when I wrote the screens I used the element that fit better the size property).  But I suppose those ones are Subtitles instead.

jmbidev: @nahudalla  @MPucheta

nahudalla (approved):

PR #64, opened 2020-01-15
Animate progress button on DescriptionRequest screen

jmbidev:
Progress bar in progress button now is animated.

### DIFF
--- a/src/components/buttons/ProgressButton.js
+++ b/src/components/buttons/ProgressButton.js
@@ -1,21 +1,36 @@
-import React, { useState } from 'react'
-import { View, StyleSheet } from 'react-native'
+import React, { useState, useEffect, useRef } from 'react'
+import { View, StyleSheet, Animated } from 'react-native'
 import { PrimaryButton, GrayButton } from '~/components'
 import { theme } from '~/constants'
 
 const ProgressBar = ({ progress }) => {
+  const progressAnimated = useRef(new Animated.Value(0))
+
+  useEffect(() => {
+    Animated.timing(progressAnimated.current, {
+      toValue: progress,
+      duration: 300
+    }).start()
+  }, [progress])
+
+  const width = progressAnimated.current.interpolate({
+    inputRange: [0, 100],
+    outputRange: ['0%', '100%'],
+    extrapolate: 'clamp'
+  })
+
   const styles = StyleSheet.create({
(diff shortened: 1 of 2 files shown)


src/components/buttons/ProgressButton.js
```diff
@@ -1,21 +1,36 @@
...
 
 const ProgressBar = ({ progress }) => {
+  const progressAnimated = useRef(new Animated.Value(0))
+
+  useEffect(() => {
+    Animated.timing(progressAnimated.current, {
+      toValue: progress,
+      duration: 300
+    }).start()
+  }, [progress])
```
nahudalla: You should stop the animation before starting a new one. The `useEffect` API allows to do that easily:
```javascript
useEffect(() => {
  const animation = Animated.timing(progressAnimated.current, {
    toValue: progress,
    duration: 300
  })
  
  animation.start()

  return () => animation.stop()
}, [progress])
```

nahudalla (changes requested):

src/components/buttons/ProgressButton.js
```diff
@@ -1,21 +1,36 @@
-import React, { useState } from 'react'
-import { View, StyleSheet } from 'react-native'
+import React, { useState, useEffect, useRef } from 'react'
+import { View, StyleSheet, Animated } from 'react-native'
```
MPucheta, line 2: If this works there is not need to change it now, but `Animated` from `react-native` is not the best way to do animations, as it requires communication between the JS thread and the native one. If the JS thread is busy, a frame drop can be seen in the animation. Use `useNativeDriver: true` option for now.

`Animated` from `react-native-animatable` or
`Animated` from `react-native-reanimated`

Are preferred options, but they are more time consuming to implement, so I think this is ok for now.

src/components/buttons/ProgressButton.js
jmbidev: For what I understand there's no need to stop animations in this case. Animations get interrupted when others animations take place. Otherwise I think that would be visible in the progressBar.

src/components/buttons/ProgressButton.js
jmbidev: Property useNativeDriver cant be set to true in this case. Native Driver doesn't support width property.

nahudalla (approved):

PR #74, opened 2020-01-22
MP/ Adapt system status bar to styleguide

MPucheta:
- Also added new generic components for App status bar and specific status bar for DetailedDiagnose

preview

### DIFF
--- a/src/components/statusBars/StatusBar.js
+++ b/src/components/statusBars/StatusBar.js
@@ -9,9 +9,9 @@ const MyStatusBar = (props) => {
     ...restProps
   } = props
 
-  let contentColor
-  contentColor = (lightContent) ? 'light-content' : 'default'
-  contentColor = (darkContent) ? 'dark-content' : 'default'
+  let contentColor = 'default'
+  if (lightContent) contentColor = 'light-content'
+  if (darkContent) contentColor = 'dark-content'
 
   return (
     <View style={style}>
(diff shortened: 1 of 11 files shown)


src/components/statusBars/StatusBar.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import { StatusBar, View } from 'react-native'
+
+const MyStatusBar = (props) => {
```
nahudalla: :thinking: I'd suggest importing React's `StatusBar` with another name and use the correct name for our component. This name is shown on the logs when debugging.
```javascript
import { StatusBar as ReactStatusBar, View } from 'react-native'

const StatusBar = (props) => {
  // ...
}
// ...
```

src/navigationService/index.js
```diff
@@ -27,6 +27,14 @@ export const ForceRerenderOnNavigation = ({ resetStateFunction }) => (
   />
 )
 
+export const ForceCleanUpOnScreenLeave = ({ cleanUpFunction }) => (
+  <NavigationEvents
+    onWillBlur={() => {
+      cleanUpFunction()
+    }}
```
nahudalla: Is this lambda really necessary? If `cleanUpFunction` doesn't expect anything as parameters, then you can just pass that funciton as the event handler: `onWillBlur={cleanUpFunction}`

nahudalla (changes requested):

nahudalla (approved):

PR #75, opened 2020-01-22
JB/ Fix progress stuck at 100% and disable back button during upload.

jmbidev:
Fixed a few things: 
  - Upload progress value addition was incorrect
  - Changed all screens to use navigation push & pop
  - Back button blocked/hided during upload of request.
  - FinishRequest's android's back button now redirects to top instead of previous screen

### DIFF
--- a/src/screens/MainApp/DiagnoseRequest/screens/DescriptionRequest/index.js
+++ b/src/screens/MainApp/DiagnoseRequest/screens/DescriptionRequest/index.js
@@ -74,7 +74,7 @@ const DescriptionRequest = ({ navigation }) => {
       await updateDatabase(uploadedImages)
       setUploadProgress(100)
       setUploading(false)
-      navigation.push('FinishRequest')
+      navigation.navigate('FinishRequest')
     } else {
       Alert.alert('No se han podido subir las imagenes. Intente nuevamente o seleccione otras imágenes.')
     }
--- a/src/screens/MainApp/DiagnoseRequest/screens/DescriptionRequest/index.js
+++ b/src/screens/MainApp/DiagnoseRequest/screens/DescriptionRequest/index.js
@@ -27,9 +27,10 @@ const DescriptionRequest = ({ navigation }) => {
       navigation.setParams({
         hideBack: true
       })
-      const handler = BackHandler.addEventListener('hardwareBackPress', () => {
-        return true
-      })
+      const handler = BackHandler.addEventListener(
+        'hardwareBackPress',
+        () => true
+      )
       return () => {
         handler.remove()
       }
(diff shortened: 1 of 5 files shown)


src/screens/MainApp/DiagnoseRequest/screens/DescriptionRequest/index.js
```diff
@@ -16,18 +16,33 @@ const DescriptionRequest = ({ navigation }) => {
...
   const imagesUris = navigation.getParam('images', [])
 
+  React.useEffect(() => {
+    if (uploading) {
+      navigation.setParams({
+        hideBack: true
+      })
+      const handler = BackHandler.addEventListener('hardwareBackPress', () => {
+        return true
+      })
```
nahudalla: Do you think this notation is more concise and readable?
```javascript
const handler = BackHandler.addEventListener(
  'hardwareBackPress',
  () => true
)
```

src/screens/MainApp/DiagnoseRequest/screens/DescriptionRequest/index.js
```diff
@@ -51,14 +66,15 @@ const DescriptionRequest = ({ navigation }) => {
         const uid = await StorageService.uploadImageAndReturnReference(uri, onProgress)
         uploadedImages.push(uid)
       } catch (error) {
-        setUploadProgress(v => v + 90 / imagesUris.length)
       }
+      currentProgress.current += 90 / imagesUris.length
```
nahudalla: If you're not handling the error but you need to do something whether an error occurred or not, use a `finally` block:
```javascript
try {
  const uid = await StorageService.uploadImageAndReturnReference(uri, onProgress)
  uploadedImages.push(uid)
} finally {
  currentProgress.current += 90 / imagesUris.length
}
```

src/storageService/index.js
```diff
@@ -19,9 +19,7 @@ export const uploadImageAndReturnReference = async (imageURI, onProgress) => {
     const blob = await BlobService.buildBlobFromURI(imageURI)
     const uuid = uuidv4()
     const imageRef = firebase.storage().ref('images').child(uuid).put(blob, metadata)
-    imageRef.on('state_changed', onProgress)
-    await imageRef
-    return uuid
+    return new Promise((resolve, reject) => imageRef.on('state_changed', onProgress, reject, () => resolve(uuid)))
```
nahudalla: :thinking: this line is a bit too long IMO. What do you think writing it like this?
```javascript
return new Promise((resolve, reject) => {
  imageRef.on(
    'state_changed',
    onProgress,
    reject,
    () => resolve(uuid)
  )
})
```

nahudalla (changes requested):

nahudalla (approved):

PR #76, opened 2020-01-23
JB/ Remove user's diagnoses script

jmbidev:
This script removes all user's diagnoses and images from the database (old and new schema).
This script needs both schemas updated to run.  
When new schema is applied it will be necessary to make little changes on the script.

To run the script use:

`node removeUserRequests <userEmail>` from scripts folder

### DIFF
--- a/scripts/removeUserRequests.js
+++ b/scripts/removeUserRequests.js
@@ -1,0 +1,106 @@
+const admin = require('firebase-admin')
+const { initFirebaseAdmin } = require('./utils/initFirebaseAdmin.js')
+
+if (process.argv.length !== 3) {
+  console.error('Invalid number of params')
+  console.log('Usage: node removeUserRequests.js <email>')
+  process.exit(1)
+}
+
+const mail = process.argv[2]
+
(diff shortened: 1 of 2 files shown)


scripts/removeUserRequests.js
```diff
@@ -0,0 +1,106 @@
...
+
+initFirebaseAdmin()
+
+const bucket = admin.storage().bucket()
+
+const asyncForEach = async (iterable, asyncCallback) => {
+  for (let i = 0; i < iterable.length; i++) {
+    await asyncCallback(iterable[i])
+  }
+}
```
nahudalla, line 20: JavaScript has the concept of an `iterable` and it's not always an Array. If you're calling the variable `iterable`, then use a `for..of` to iterate over it. Otherwise, make it explicit that you are expecting an array.
```javascript
const asyncForEach = async (iterable, asyncCallback) => {
  for (const item of iterable) {
    await asyncCallback(item)
  }
}
```

nahudalla (approved):

PR #77, opened 2020-01-23
MP/ Add cloud function to set or increment amountOfAnswers

MPucheta:
- This is triggered on each response creation for a request

- If `amountOfAnswers` does not exist, it is set to 1 (as a new response triggers the initial value)

- if `amountOfAnswers` exists, then it gets incremented by 1

An admin or user comment triggers this functions, as those should create a new response.

image


nahudalla (approved):
Excellent, I don't remember if we have a task for this, but what happens with requests that already exist?

PR #78, opened 2020-01-23
MP/ Add script to set amountOfAnswers for existing answered requests

MPucheta:
- This uses collectionGroup queries, every document inside a subcollection 'requests' enters the query and gets filtered if it has an 'answered' field set as true. For each on of those, the amountOfAnswers is set to have a value of 1.

- **A new composite index is needed for this, executing this script without setting up the index will fail execution and show the link to set up the index in console**

image

- **To run this script, refer to documentation regarding setup and execution for scripts**

### DIFF
--- a/scripts/setAmountOfAnswersInExistingAnsweredRequests.js
+++ b/scripts/setAmountOfAnswersInExistingAnsweredRequests.js
@@ -15,8 +15,14 @@ const setAmountOfAnswersInExistingAnsweredRequests = async () => {
     await Promise.all(
       snapshots.docs.map(async (doc) => {
         console.log(`Setting amount of answers for ${doc.id}`)
+
+        const responses = await admin
+          .firestore()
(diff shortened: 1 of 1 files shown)


scripts/setAmountOfAnswersInExistingAnsweredRequests.js
```diff
@@ -0,0 +1,30 @@
...
+      .where('answered', '==', true)
+
+    const snapshots = await answeredRequests.get()
+
+    await Promise.all(
+      snapshots.docs.map(async (doc) => {
+        console.log(`Setting amount of answers for ${doc.id}`)
+        await doc.ref.update({
+          amountOfAnswers: 1
+        })
```
nahudalla: What do you think about actually counting the amount of answers?

nahudalla (approved):

PR #83, opened 2020-01-27
JB/ Add new response fields on detailed diagnose screen

jmbidev:
Input and button floating at the bottom of the screen

image

### DIFF
--- a/src/databaseService/index.js
+++ b/src/databaseService/index.js
@@ -162,7 +162,6 @@ export const addDiagnoseResponse = async (diagnoseUID, response) => {
 
     const responseData = {
       answer: response,
-      answeredBy: username,
       answeredByUID: userUID
     }
 
--- a/src/databaseService/index.js
+++ b/src/databaseService/index.js
@@ -154,6 +154,26 @@ export const addNewUserData = async (userUID, userData) => {
   }
 }
 
+export const addDiagnoseResponse = async (diagnoseUID, response) => {
+  try {
+    const userUID = AuthenticationService.getCurrentUserUID()
+    const { username } = await get(`users/${userUID}`)
+    const responsePath = `users/${userUID}/requests/${diagnoseUID}/responses`
+
+    const responseData = {
+      answer: response,
+      answeredBy: username,
+      answeredByUID: userUID
+    }
+
+    const responseReference = await add(responsePath, responseData)
+    const responseFetch = await responseReference.get()
+    return responseFetch.data()
+  } catch (error) {
+    throw new DatabaseError(error.message)
+  }
+}
+
 export default {
   get,
   update,
--- a/src/databaseService/index.js
+++ b/src/databaseService/index.js
@@ -168,6 +168,7 @@ export const addDiagnoseResponse = async (diagnoseUID, response) => {
 
     const responseReference = await add(responsePath, responseData)
     const responseFetch = await responseReference.get()
+
     return responseFetch.data()
   } catch (error) {
     throw new DatabaseError(error.message)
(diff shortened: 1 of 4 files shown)


src/databaseService/index.js
```diff
@@ -154,6 +154,26 @@ export const addNewUserData = async (userUID, userData) => {
...
+  try {
+    const userUID = AuthenticationService.getCurrentUserUID()
+    const { username } = await get(`users/${userUID}`)
+    const responsePath = `users/${userUID}/requests/${diagnoseUID}/responses`
+
+    const responseData = {
+      answer: response,
+      answeredBy: username,
+      answeredByUID: userUID
+    }
```
nahudalla: Can we remove the `answeredBy` field on user responses and query its username by `answeredByUID` when displaying the answers? If we allow the users to change their usernames in the future, this will be a problem...

cc @MPucheta

src/databaseService/index.js
```diff
@@ -154,6 +154,26 @@ export const addNewUserData = async (userUID, userData) => {
...
+
+    const responseReference = await add(responsePath, responseData)
+    const responseFetch = await responseReference.get()
+    return responseFetch.data()
```
nahudalla, line 171: Please add an empty line before `return` statements.

src/screens/MainApp/DetailedDiagnose/components/NewComment/index.js
```diff
@@ -0,0 +1,42 @@
...
+        placeholderTextColor={theme.colors.gray}
+      />
+      {sendEnabled &&
+        <PrimaryButton onPress={handlePress}>
+          <Description white>Enviar</Description>
+        </PrimaryButton>}
+      {!sendEnabled &&
+        <GrayButton>
+          <Description white>Enviar</Description>
+        </GrayButton>}
```
nahudalla, line 37: I think this is a common pattern that is repeated a lot on the whole app. What do you think about adding a `disabled` boolean property to `PrimaryButton`? Internally, if `disabled` is `true`, it would render as a `GrayButton` and remove the `onPress` callback.

cc @MPucheta

src/screens/MainApp/DetailedDiagnose/components/NewComment/styles.js
```diff
@@ -0,0 +1,21 @@
+import { StyleSheet } from 'react-native'
+import { theme } from '~/constants'
+import { verticalScale, moderateScale } from 'react-native-size-matters/extend'
+
+const styles = StyleSheet.create({
+  container: {
+  },
```
nahudalla: Why this empty class?

nahudalla (changes requested):

src/databaseService/index.js
jmbidev: I think that field is used because admin users has nick name @nahudalla

src/screens/MainApp/DetailedDiagnose/components/NewComment/index.js
jmbidev: I think it is a good idea, I also think that there won't be disabled buttons with other color than gray.

jmbidev: I'm gonna make a change to shorten this code

MPucheta: It's a good idea. I thought about creating a "ToggleButton" but this solution is better.

src/databaseService/index.js
MPucheta: Indeed, this is a problem as adminUID do not correlate to nick name. 

We could implement the same logic for all if we create an admin's account with a real username (or change the actual ones to match those usernames)

nahudalla: I meant removing the field **only for user responses**. A response _can_ have both fields, but `answeredBy` would only be used for an Admin response. If `answeredBy` is not present, then the username would be loaded using `answeredByUID`.

MPucheta (approved):

nahudalla (approved):

PR #84, opened 2020-01-27
MP/ Add a chat-like list to show ALL responses

MPucheta:
image

### DIFF
--- a/admin-src/DiagnoseInfo/components/Response/index.js
+++ b/admin-src/DiagnoseInfo/components/Response/index.js
@@ -1,21 +1,19 @@
 import React from 'react'
-import './styles.css'
+import classnames from 'classnames'
+import classes from './styles.scss'
 
 const Response = ({ text, date, isAdmin = false }) => {
-  let messageClassname = 'response'
-
-  if (isAdmin) {
-    messageClassname += ' adminResponse'
-  } else {
-    messageClassname += ' userResponse'
-  }
+  const messageClassname = classnames(classes.response, {
+    [classes.adminResponse]: isAdmin,
+    [classes.userResponse]: !isAdmin
+  })
 
   return (
     <div className={messageClassname}>
-      <div className='text'>
+      <div className={classes.text}>
         {text}
       </div>
-      <div className='date'>
+      <div className={classes.date}>
         {date}
       </div>
     </div>
--- a/admin-src/DiagnoseInfo/components/Response/index.js
+++ b/admin-src/DiagnoseInfo/components/Response/index.js
@@ -1,0 +1,25 @@
+import React from 'react'
+import './styles.css'
+
+const Response = ({ text, date, isAdmin = false }) => {
+  let messageClassname = 'response'
+
+  if (isAdmin) {
+    messageClassname += ' adminResponse'
+  } else {
+    messageClassname += ' userResponse'
+  }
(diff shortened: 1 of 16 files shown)


admin-src/DiagnoseInfo/components/Response/index.js
```diff
@@ -0,0 +1,25 @@
...
+import './styles.css'
+
+const Response = ({ text, date, isAdmin = false }) => {
+  let messageClassname = 'response'
+
+  if (isAdmin) {
+    messageClassname += ' adminResponse'
+  } else {
+    messageClassname += ' userResponse'
+  }
```
nahudalla: There's a library called `classnames` to do exactly this:
```javascript
import classnames from 'classnames'

const messageClassname = classnames('response', {
  adminResponse: isAdmin,
  userResponse: !isAdmin
})
```

admin-src/DiagnoseInfo/components/Response/index.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import './styles.css'
```
nahudalla: Please configure **PostCSS Modules** (it should be only one config file on the root and Parcel handles it automatically).

With PostCSS Modules you can _import_ the classes and their names get mangled to avoid duplicates in the final bundle:

```javascript
import classes from './styles.css'

//later you can use the classes like this:
<div className={classes.someCSSClass} />
```

admin-src/DiagnoseInfo/components/Response/styles.css
```diff
@@ -0,0 +1,34 @@
...
+  width: 50%;
+  margin: 1em;
+  padding: 0.5em;
+  display: flex;
+  overflow: hidden;
+  justify-content: center;
+  border-radius: 0.5em;
+  -webkit-box-shadow: 0px 0px 6px 0px rgba(148,148,148,1);
+-moz-box-shadow: 0px 0px 6px 0px rgba(148,148,148,1);
+box-shadow: 0px 0px 6px 0px rgba(148,148,148,1);
```
nahudalla: No need for prefixed properties. Change the extension of the file to `.scss`. It'll be a SASS stylesheet. SASS is _almost_ like CSS but gets compiled before Parcel adds it to the bundle. Sass handles these prefixes automatically.

admin-src/DiagnoseInfo/components/Responses/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+  useEffect(() => {
+    const scrollToBottom = () => {
+      scrollToBottomRef.current.scrollTop = scrollToBottomRef.current.scrollHeight
+    }
+
+    if (scrollToBottomRef.current) {
+      scrollToBottom()
+    }
+  }, [scrollToBottomRef])
```
nahudalla: This will execute only once. A ref is an object *that never changes*, only its `current` property changes. You can pass a function to a `ref` prop. The function will get called with a reference to the element. Use that to scroll to the bottom on initialization. Otherwise, consider executing this effect every time `responses` changes.

```javascript
<div ref={ref => { /* do something with ref */ }} />
```

nahudalla (changes requested):

nahudalla (approved):

PR #87, opened 2020-01-29
JB/ admin page diagnose sections

jmbidev:
Set the amount of days before that stale status to 10
image

### DIFF
--- a/admin-src/DiagnoseResponse/index.js
+++ b/admin-src/DiagnoseResponse/index.js
@@ -8,52 +8,102 @@ import { firebaseTimestampToMoment } from '../utils/date'
 import '../stylesheets/admin.css'
 import { getUserUIDFromDiagnoseRef } from '../utils/diagnose'
 
-export const DiagnoseResponse = () => {
+const STALE_STATUS_AFTER_DAYS = 10
+
+export const DiagnoseResponse = ({ filter }) => {
   const [currentDiagnose, setCurrentDiagnose] = useState(null)
   const [diagnoses, setDiagnoses] = useState([])
   const [isSubmitting, setIsSubmitting] = useState(false)
 
-  useEffect(() => {
-    const attachQueryListenerForUnansweredDiagnoses = () => {
-      firebase
-        .firestore()
-        .collectionGroup('requests')
-        .where('amountOfAnswers', '==', 0)
-        .onSnapshot(async (querySnapshot) => {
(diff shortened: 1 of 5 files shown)


admin-src/DiagnoseResponse/index.js
```diff
@@ -8,52 +8,107 @@ import { firebaseTimestampToMoment } from '../utils/date'
...
+      } else if (filter === 'stale') {
+        staleQuery()
+      } else if (filter === 'solved') {
+        solvedQuery()
+      }
     }
 
-    attachQueryListenerForUnansweredDiagnoses()
+    attachQueryListener()
   }, [])
```
nahudalla: I think it would be better if `filter` is actually the function that filters the data, not a string. It's not a good design choice to use strings to select options. It's better to make the component open so that any filter can be applied.

nahudalla (changes requested):

admin-src/DiagnoseResponse/index.js
jmbidev: I think what is wrong is the name of the variable it should not be 'filter' but 'state' instead that's why it lead you to that. Don't forget almost all react-native components use strings in select options.

nahudalla: That React Native uses this pattern doesn't mean it's correct. You're breaking the open/closed principle here. A component should be open to extension but closed to modification. Consider what happens if we need to change/remove/add a filter:
- **With your solution:** we would need to _modify the component_ to change/remove/add the filter.
- **With my suggestion:** you only have to create the new filter and pass that as a prop. This is called dependency inversion/injection if I remember correctly.

cc @MPucheta what do you think?

MPucheta: I agree with Nahuel's suggestion. These "strings" options are almost always solved by using some form of polymorphism too (I do not like explicit inheritance hierarchy, but you could have all these filter follow the same interface and achieve the same effect, for example, all of them receive a snapshot and a filter) I would also try to separate all those filters and queries into separate files.

jmbidev: I agree with your comments, what you are suggesting is obviously a better solution object oriented. It also would be better to not pass any filter but the diagnose objects instead. I just applied this solution for the given delivery date.  I'll make the changes

nahudalla (approved):

PR #88, opened 2020-01-29
JB/ push notification redirect onpress

jmbidev:
Opening a notification inside or outside the app now redirects to MyDiagnoses page.


src/App.js
```diff
@@ -14,6 +14,8 @@ import Onboarding from './screens/Onboarding'
 import MainApp from './screens/MainApp'
 import PrivacyPolicy from './screens/PrivacyPolicy'
 import { buildStackNavigator } from '~/components/StackNavigator'
+import notificationService from '~/notificationService'
+import navigationService from './navigationService'
```
nahudalla: This was already imported.

src/notificationService/index.js
```diff
@@ -0,0 +1,57 @@
...
+  return _notifications.shift()
+}
+
+export default {
+  configure,
+  setNotificationHandler,
+  enableHandler,
+  disableHandler,
+  popInitialNotification
+}
```
nahudalla: Don't export objects like this as default. You have already exported all these functions and they can be imported as an object using `import * as obj from 'module'`.

nahudalla (changes requested):

nahudalla (approved):

PR #91, opened 2020-01-31
JB/ Add No Connection screen to redirect when there is no connection

jmbidev:
This change adds a NoConnection screen to redirect when there is no internet connection.
For now it's shown when the user opens the app and there is no internet.
There would be necessary to redirect to this screen from others screens or it is posible to set an event listener that redirects when it detects that there is no internet connection.

Screenshot_1580486986


nahudalla (approved):
@jmbidev please resolve the conflicts.

PR #93, opened 2020-02-03
JB/ Improve MyDiagnoses rendering

jmbidev:
Changed the way diagnoses are retrieved and the screen updated.

Now we listen to database changes instead of download all diagnoses every time we navigate to the screen.


src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -77,34 +72,29 @@ const MyDiagnoses = () => {
...
+          onScrollBeginDrag={() => setScrolling(true)}
+          onScrollEndDrag={({ nativeEvent }) => {
+            const snapToTopIfScrollingNearTheFirstElement = () => {
+              if (nativeEvent.contentOffset.y < OFFSET_THRESHOLD_TO_SNAP_FLATLIST) {
+                setScrolling(false)
+                flatListRef.current.scrollToIndex({ animated: true, index: 0 })
+              } else {
+                setScrolling(true)
               }
+            }
```
nahudalla: Why is this function defined if it's called immediately?

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -1,74 +1,69 @@
...
+const buildDiagnose = (doc) => {
+  const diagnose = doc.data({ serverTimestamps: 'estimate' })
+  diagnose.thumbnail = diagnose.imageReferences[0]
+  return diagnose
+}
+
 const MyDiagnoses = () => {
   const flatListRef = useRef()
-  const [diagnoses, setDiagnoses] = useState(null)
+  const diagnoses = useRef([])
```
nahudalla: Why are you using a reference here instead of a state? If you need this value for rendering, then you _have to_ use `useState`. References don't cause re-renders of the component. All the values you use for rendering must be immutable.

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -77,34 +72,29 @@ const MyDiagnoses = () => {
...
-                  flatListRef.current.scrollToIndex({ animated: true, index: 0 })
-                  setScrolling(false)
-                } else {
-                  setScrolling(true)
-                }
+        <FlatList
+          ref={flatListRef}
+          data={diagnoses.current}
+          showsVerticalScrollIndicator={false}
+          renderItem={({ item }) => item.renderedDiagnose}
```
nahudalla: I don't like much this pattern of pre-rendering. I don't know if it plays nicely with React's internals (reconciliation in particular, automatic memory management [we're keeping all the diagnoses rendered in-memory and preventing garbage-collection]). Instead of a function `renderDiagnose` and storing the diagnose here, you could just have a component `DiagnoseListItem` that receives an `item` prop, and pass that component here. Example:
```javascript
// in this function:
  <FlatList
    // ...
   renderItem={Diagnose}
   // ...
  />
  // ...

function Diagnose ({ item }) {
  return // render the diagnose here
}
```

What do you guys think? cc @MPucheta

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -1,74 +1,69 @@
...
-    try {
-      const downloadedData = await DatabaseService.getDiagnosesFromCurrentUser()
-
-      const downloadedDataWithThumbnail = await Promise.all(
-        downloadedData.map(async (diagnose, index) => {
-          diagnose.thumbnail = await getURL(diagnose.imageReferences[0])
-          return diagnose
-        })
-      )
+          if (changeType === 'added') {
```
nahudalla: What do you think of replacing this `if...else...if...else` chain with a `switch` statement and moving the logic for each case to its own function?
```javascript
switch (changeType) {
  case 'added': setDiagnoses(diagnoses => processNewDiagnose(previousDiagnoses, diagnose)); break
  case 'modified': setDiagnoses(diagnoses => processModifiedDiagnose(previousDiagnoses, diagnose)); break
  case 'removed': setDiagnoses(diagnoses => processRemovedDiagnose(previousDiagnoses, diagnose)); break
  default: throw new Error('Error processing snapshot from Firestore')
}

// All the processXXXDiagnose functions return a new array with the diagnoses
```

nahudalla (changes requested):

src/screens/MainApp/MyDiagnoses/index.js
nahudalla: If it's because of readability, it can be improved by using a function as a predicate of the `if` statement:
```javascript
              if (isScrollWithinSnapThreshold(nativeEvent)) {
                setScrolling(false)
                flatListRef.current.scrollToIndex({ animated: true, index: 0 })
              } else {
                setScrolling(true)
              }
```

jmbidev: I tried not to change too much of what was already there, I think we can change that. @MPucheta  what do you think?

src/screens/MainApp/MyDiagnoses/index.js
jmbidev: Mmm. I think we can save some time of rendering if we actually save pre-rendered objects but obviously it uses more memory. Right now there is no big impact on memory because there are few diagnoses.  In a future having +50 diagnoses could change this. I think we should measure it to get real thoughts of what is better...

src/screens/MainApp/MyDiagnoses/index.js
jmbidev: Definitly it's cleaner

jmbidev: default clause it's not needed

src/screens/MainApp/MyDiagnoses/index.js
jmbidev: Seems that pre-rendering like this doesn't save any time on rendering so I'm making the change.

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -18,9 +18,28 @@ const buildDiagnose = (doc) => {
...
+  const index = diagnoses.findIndex(diagnose => diagnose.id === doc.id)
+  diagnoses[index] = { builtDiagnose, id: doc.id }
+  return diagnoses
+}
+
+const processRemovedDiagnose = (diagnoses, doc) => {
+  const index = diagnoses.findIndex(diagnose => diagnose.id === doc.id)
+  diagnoses.splice(index, 1)
+  return diagnoses
+}
```
nahudalla: - Please add an empty line after constant declarations and before return statements
- These **functions are not pure**. They must not mutate the parameters. Create new arrays and return that (using spread operator to make a copy and `filter` to remove an item).
- Remember that for us, a modified diagnose is simply its deletion and immediate addition to the array. So, you can reuse the other two functions for that one.

nahudalla (changes requested):

src/screens/MainApp/MyDiagnoses/index.js
nahudalla: The [**boy scout rule**](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html) tells us to leave the place better than we found it. If you find code that could be written better and it's a minor change, then improve it.

nahudalla:
@jmbidev also resolve the conflicts please

nahudalla (approved):

PR #94, opened 2020-02-04
MP/ Change design for floating message field

MPucheta:
- Replace NewComment with the new component

- This prompts the creation of a SvgButton component and a change in MyDiagnoses to remove tabBarVisibility in screens from that stack

image

image


src/screens/MainApp/DetailedDiagnose/components/NewComment/index.js
```diff
@@ -1,40 +1,95 @@
...
+      'keyboardDidHide',
+      keyboardDidHide
+    )
+
+    const cleanUp = () => {
+      keyboardDidShowListener.remove()
+      keyboardDidHideListener.remove()
+    }
+
+    return cleanUp
```
nahudalla: :thinking: I think creating functions to immediately use them creates unnecessary noise. Every should programmer knows what `addListener` means in any context and programming environment, so it should be obvious that the function you're passing as second parameter is the event handler. Also, every programmer using React should know its API. It should be obvious that if you return a function, it's a cleanup function. This is much cleaner and shorter (and in turn, easier to read) IMO:
```javascript
useEffect(() => {
    const keyboardDidShowListener = Keyboard.addListener(
      'keyboardDidShow',
      () => setCommentPadding(INPUT_PADDING_ON_KEYBOARD_OPENED)
    )

    const keyboardDidHideListener = Keyboard.addListener(
      'keyboardDidHide',
      () => setCommentPadding(INPUT_PADDING_STATIC)
    )

    return () => {
      keyboardDidShowListener.remove()
      keyboardDidHideListener.remove()
    }
}, [])
```

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -113,9 +113,31 @@ MyDiagnoses.navigationOptions = () => ({
...
+      if (includes(allowedRoutesToShowTabBar, route.routeName)) {
+        tabBarVisible = true
+      } else {
+        tabBarVisible = false
+      }
+    })
+  }
+
+  return { tabBarVisible }
+}
```
nahudalla: - The constant `allowedRoutesToShowTabBar` should be in the top-level, otherwise you'd be creating a new array on each call unnecessarily.
- Never use `map` to iterate over an array, it's meant to **modify** the contents of the array. Use `forEach` for iterating, it doesn't create a new array.
- This code can be written in a much shorter and readable way. I think it's the same behavior?
```javascript
import { isEmpty } from 'lodash'
// ...
const allowedRoutesToShowTabBar = ['MyDiagnoses']
// ...
MyDiagnosesStack.navigationOptions = ({ navigation }) => {
  if (navigation.state.routes.length <= 1) return {}

  const routesInHistoryWithTabBar = intersect(navigation.state.routes, allowedRoutesToShowTabBar)
  
  return { tabBarVisible: !isEmpty(routesInHistoryWithTabBar) }
}
```

nahudalla (changes requested):

src/screens/MainApp/MyDiagnoses/index.js
MPucheta: I changed the code to reflect this. This should be executed once though. It's a change in a static property of the component. Nevertheless, once we stop modifying this component to add new features we should move this into a separate module like the one in `DiagnoseRequest`

nahudalla (approved):

PR #96, opened 2020-02-05
JB/ Add Solve and Remove requests feature

jmbidev:
What is done:

- Added removedAt and isSolved fields to database
- Added script to set all diagnoses to not removed `removedAt == 0` (needed to successfully fetch diagnoses not removed). This script must be executed only once.
- Fixed error with header in DetailedDiagnose screen (zIndex)
- Added index to database to fetch not removed diagnoses and updated the query
- Added SolutionRequest & FinishScreen  (solve diagnose process).

Screenshot_1580920352
Screenshot_1580920363
Screenshot_1580920355
Screenshot_1580920373
Screenshot_1580920393
Screenshot_1580920406
Screenshot_1580920419


src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -110,21 +113,25 @@ MyDiagnoses.navigationOptions = () => ({
...
+  if (navigation.state.routes.length > 1) {
+    navigation.state.routes.map(route => {
+      tabBarVisible = includes(allowedRoutesToShowTabBar, route.routeName)
+    })
+  }
 
-  return { tabBarVisible: !isEmpty(routesInHistoryWithTabBar) }
+  return { tabBarVisible }
 }
-
```
MPucheta: You are reverting changes. The code with `intersection` is the one working, you should only add `FinishScreen` to `allowedRoutesToShowTabBar`

src/databaseService/index.js
```diff
@@ -91,7 +91,9 @@ export const addDiagnose = async (imageReferences, text) => {
       user,
       text,
       imageReferences,
-      amountOfAnswers: 0
+      amountOfAnswers: 0,
+      removedAt: 0,
+      isSolved: false
```
MPucheta, line 96: This probably would need another script to set `isSolved` in existing requests.

@nahudalla

src/screens/MainApp/MyDiagnoses/screens/FinishScreen/index.js
```diff
@@ -0,0 +1,52 @@
...
+const FinishScreen = ({ navigation }) => {
+  AnalyticsService.setCurrentScreenName('Finish Screen')
+
+  React.useEffect(() => {
+    const handler = BackHandler.addEventListener('hardwareBackPress', () => {
+      navigation.popToTop()
+      return true
+    })
+    return () => handler.remove()
+  }, [])
```
MPucheta: Please import `useEffect` and add a line before the return

src/screens/MainApp/MyDiagnoses/screens/SolutionRequest/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+      navigation.navigate('FinishScreen')
+    } catch (error) {
+      console.log(error)
+      Alert.alert(
+        'Error',
+        'No se pudo realizar la operación. Por favor, intente nuevamente más tarde.'
+      )
+      setUploading(false)
+      setInputEnabled(true)
```
MPucheta: Please remove the `console.log` and a blank before `setUploading` and `setInputEnabled`

src/screens/MainApp/MyDiagnoses/screens/SolutionRequest/index.js
```diff
@@ -0,0 +1,131 @@
...
+      })
+      const handler = BackHandler.addEventListener(
+        'hardwareBackPress',
+        () => true
+      )
+      return () => {
+        handler.remove()
+      }
+    }
+  }, [uploading])
```
MPucheta: Please import `UseEffect` and add a blank before the `return`

MPucheta (dismissed):
Nice job! I left a couple of comments, i'm requesting changes as you are reverting some code back regarding the navigation options in `MyDiagnoses`. Nahuel asked me to replace the code as `map` was not the correct way to do this, and overall the new code was more compact and less verbose.

src/databaseService/index.js
jmbidev: Yes probably in a future it will be needed if we want to query for that value. But it is not needed for this feature.

src/screens/MainApp/MyDiagnoses/index.js
jmbidev: I tested it and it did not work (FinishScreen did not have the tab bar). Thats why I changed it back.
@MPucheta

jmbidev: It doesn't work because you need to access `route.routeName` to retrieve the name of the route.

jmbidev: Changed it to be more readable

src/databaseService/index.js
```diff
@@ -184,14 +186,37 @@ export const getUsernameByUID = async (userUID) => {
...
+  try {
+    const userUID = AuthenticationService.getCurrentUserUID()
+    const removedData = {
+      removedAt: (isRemoved) ? firebase.firestore.FieldValue.serverTimestamp() : 0
+    }
+
+    return update(`users/${userUID}/requests/${diagnoseUID}`, removedData)
+  } catch (error) {
+    throw new DatabaseError(error.message)
+  }
```
nahudalla: You're not `await`ing the promise. It won't `throw` the error so the `try/catch` surrounding this won't catch the error.

src/databaseService/index.js
```diff
@@ -184,14 +186,37 @@ export const getUsernameByUID = async (userUID) => {
...
+}
+
+export const setDiagnoseSolvedMark = async (diagnoseUID, isSolved) => {
+  try {
+    const userUID = AuthenticationService.getCurrentUserUID()
+
+    return update(`users/${userUID}/requests/${diagnoseUID}`, { isSolved: isSolved === true })
+  } catch (error) {
+    throw new DatabaseError(error.message)
+  }
```
nahudalla: - You're not `await`ing the promise. It won't `throw` the error so the `try/catch` surrounding this won't catch the error.
- I think `{ isSolved: isSolved ? true : false }` is better for the user of this function. This way any [truthy value](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) can be used.

src/screens/MainApp/DetailedDiagnose/components/HeaderForScrolling/index.js
```diff
@@ -9,22 +11,27 @@ import styles from './styles'
 
 const TRANSITION_DURATION = 100
 
-const HeaderForScrolling = ({ show, onGoBack, date, commentCount }) => {
+const HeaderForScrolling = ({ show, onGoBack, date, commentCount, solved, onSolveOrOpen }) => {
+  const [hide, setHide] = useState(true)
```
nahudalla: `hide` is a verb. `hidden` would be the sustantive for this variable. I also like `isHidden`. (I think Clean Code suggests to add `is` before boolean variables. cc @MPucheta do you remember?)

src/screens/MainApp/DetailedDiagnose/components/HeaderForScrolling/index.js
```diff
@@ -42,6 +49,7 @@ const HeaderForScrolling = ({ show, onGoBack, date, commentCount }) => {
           data={commentCount}
           svg={comments}
         />
+        <Body style={styles.metadataAsRow} secondary onPress={onSolveOrOpen}>{solvedText}</Body>
```
nahudalla: I think it's better to have these events as separate (`onSolveTapped`, `onReopenTapped`). You're forcing all the potential users of this component to check for the `solved` value themselves, you can do it here and simplify the usage of this component.

src/screens/MainApp/DetailedDiagnose/index.js
```diff
@@ -22,10 +22,55 @@ import StatusBarOnScroll from './components/StatusBarOnScroll'
...
 import { renderResponses, renderResponse } from './renderUtilities'
 import { OFFSET_THRESHOLD_TO_CHANGE_HEADER } from './constants'
+import { useActionSheet } from '@expo/react-native-action-sheet'
 import styles from './styles'
 
+const goToSolveDiagnoseScreen = (diagnose) => {
+  NavigationService.navigate('SolutionRequest', { diagnose })
+}
+
+const reOpenDiagnose = async (diagnose) => {
```
nahudalla: `reopen` is a single word

src/screens/MainApp/DetailedDiagnose/index.js
```diff
@@ -53,6 +99,35 @@ const DetailedDiagnose = ({ navigation }) => {
...
+        cancelButtonIndex
+      },
+      buttonIndex => {
+        switch (buttonIndex) {
+          case 0: decideIfCloseOrOpen(); break
+          case 1: askForDeleteConfirmation(diagnose, navigation); break
+        }
+      }
+    )
+  }, [])
```
nahudalla: You're using `diagnose`, `decideIfCloseOrOpen`, `askForDeleteConfirmation` and `navigation` from the outer scope in this callback. You should add any variables that can change to the dependencies array. IMO, `diagnose` and perhaps `navigation` _will_ change eventually.

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -110,21 +119,23 @@ MyDiagnoses.navigationOptions = () => ({
...
 
-const allowedRoutesToShowTabBar = ['MyDiagnoses']
+const allowedRoutesToShowTabBar = ['MyDiagnoses', 'FinishScreen']
 
 MyDiagnosesStack.navigationOptions = ({ navigation }) => {
   if (navigation.state.routes.length <= 1) return {}
 
-  const routesInHistoryWithTabBar = intersection(navigation.state.routes, allowedRoutesToShowTabBar)
+  const routesInState = navigation.state.routes.map(route => route.routeName)
+  const tabBarVisible = includes(allowedRoutesToShowTabBar, routesInState.pop())
```
nahudalla: The `map` and `pop` are unnecessary. The following code is equivalent and `O(1)`:
```javascript
const routes = navigation.state.routes
const latestRouteName = routes[routes.length - 1].routeName
const tabBarVisible = includes(allowedRoutesToShowTabBar, latestRouteName)
```

src/screens/MainApp/MyDiagnoses/screens/SolutionRequest/index.js
```diff
@@ -0,0 +1,130 @@
...
+  const handleUpload = async () => {
+    setInputEnabled(false)
+    setUploading(true)
+    try {
+      await DatabaseService.addDiagnoseResponse(diagnose.id, description)
+      await DatabaseService.setDiagnoseSolvedMark(diagnose.id, true)
+
+      setDescription('')
+      setUploading(false)
+      setInputEnabled(true)
```
nahudalla: You're repeating these here and on the `catch` block. Move them to the `finally` block.

nahudalla (changes requested):
### Excellent! :ok_hand:

I left a few comments below.

src/screens/MainApp/DetailedDiagnose/index.js
```diff
@@ -208,7 +208,8 @@ const DetailedDiagnose = ({ navigation }) => {
         date={date}
         commentCount={amountOfResponses}
         solved={diagnose.isSolved}
-        onSolveOrOpen={decideIfCloseOrOpen}
+        onSolveTapped={decideIfCloseOrOpen}
+        onReopenTapped={decideIfCloseOrOpen}
```
nahudalla: Now that you separate props for each event, there's no need for a `decideIf...` function. You can have them as separate handlers:
```javascript
  const goToSolveCurrentDiagnose = useCallback(() => goToDiagnoseScreen(diagnose), [diagnose])
  const reopenCurrentDiagnose = useCallback(() => reopenDiagnose(diagnose), [diagnose])
// then in the props:
  onSolveTapped={goToSolveCurrentDiagnose}
  onReopenTapped={reopenCurrentDiagnose}
```

_These props now read as a sentence in English_ 😄

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -133,8 +133,8 @@ const allowedRoutesToShowTabBar = ['MyDiagnoses', 'FinishScreen']
 MyDiagnosesStack.navigationOptions = ({ navigation }) => {
   if (navigation.state.routes.length <= 1) return {}
 
-  const routesInState = navigation.state.routes.map(route => route.routeName)
-  const tabBarVisible = includes(allowedRoutesToShowTabBar, routesInState.pop())
+  const lastRouteInState = navigation.state.routes.pop().routeName
```
nahudalla: Do you really need to `pop` the last element? It makes this function non-pure. In my previous review I wrote an example that makes the function _pure_ on purpose. https://github.com/WeDevelop-ARG/cannabis-app/pull/96#discussion_r376469850

src/screens/MainApp/MyDiagnoses/screens/SolutionRequest/index.js
```diff
@@ -27,22 +27,22 @@ const SolutionRequest = ({ navigation }) => {
...
     try {
       await DatabaseService.addDiagnoseResponse(diagnose.id, description)
       await DatabaseService.setDiagnoseSolvedMark(diagnose.id, true)
 
       setDescription('')
-      setUploading(false)
-      setInputEnabled(true)
-
-      navigation.navigate('FinishScreen')
+      doNavigate = true
```
nahudalla, line 36: If you have to navigate away of this screen, does it matter if the navigation happens _before_ or _after_ the states are set? I think it does not. IMO the navigation _can_ be here, before setting the states. Also using boolean variables to determining if a `try/catch` succeeded is a code smell (if I remember correctly, I read that on Clean Code).

nahudalla (changes requested):

nahudalla:
@jmbidev please resolve the conflicts

nahudalla (approved):

PR #104, opened 2020-02-18
MP/ Add diagnose UUID to URL admin interface

MPucheta:
- This is a fundamental change on how we rendered the Diagnoses, and to maintain the codebase I had to change some names. 

  - `DiagnoseResponse` now is `DiagnoseList`
  - `Responses` is now `ResponsesChat`
  - There is a `DiagnosePanel` which uses all components regarding Diagnose (which is the main component rendered by URL)

- This allows us to share the diagnose URL to other admins to work on the same problem (in a database friendly way, it only fetches that request)

- It was necessary to add the userID to the url. It was impossible to fetch the diagnose otherwise, as firebase does not support this functionality (a bug regarding `collectionGroup` and `documentId()`)

- New implementation now maintains the diagnose after submitting a response, showing the admin new message and also being responsive for new messages from the user. This new implementation also implements a query snapshot with incremental loading, which is good to not hit the database more than necessary. 

preview


nahudalla:
> I implemented this to put the screen in blank after answering the diagnose as redirecting to another section to show the same request seems like a weird redirection, as one would hardly answer the same request twice in a row and would likely continue answering request from the same section. @nahudalla what do you think?

@MPucheta I explicitly asked for that change in the Trello task because it was annoying to me that the request would just disappear after I answered, and I couldn't even see my comment on the chat. Also, if we happen to engage in a conversation with the user we wouldn't want to have to open the diagnose again and again.

admin-src/DiagnoseList/index.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    return attachQueryListener()
+  }, [])
+
+  return (
+    <div className={classes.row}>
+      <ListGroup className={classes.column}>
+        {diagnoses && diagnoses.map((diagnose, index) => (
+          <Link key={index} to={match.url + `/${diagnose.userUID}/${diagnose.id}`}>
```
nahudalla: If you're using string templates, add the `match.url` within the string too.

admin-src/DiagnosePanel/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const handleSubmit = async (values) => {
+    setIsSubmitting(true)
+
+    try {
+      await writeNewResponse(values.answer, values.answeredBy, diagnose.userUID, diagnose.id)
+      setFinished(true)
+    } catch (error) {
+      alert('Unexpected problem at writing a new response, try again later.')
+      setIsSubmitting(false)
```
nahudalla, line 23: This has to be in a `finally` block.

admin-src/DiagnosePanel/index.js
```diff
@@ -0,0 +1,71 @@
...
+  const location = useLocation()
+
+  const handleSubmit = async (values) => {
+    setIsSubmitting(true)
+
+    try {
+      await writeNewResponse(values.answer, values.answeredBy, diagnose.userUID, diagnose.id)
+      setFinished(true)
+    } catch (error) {
+      alert('Unexpected problem at writing a new response, try again later.')
```
nahudalla: "when writing a new response"

admin-src/utils/routing/getStateFromLocation.js
```diff
@@ -0,0 +1 @@
+export const getStateFromLocation = (location) => location.pathname.split('/')[1] || 'new'
```
nahudalla: What is this function for? React router does this automatically. Don't try to parse URLs manually.

nahudalla (changes requested):
This will have many conflicts with Abel's PR. We'll merge his first.

nahudalla (approved):

nahudalla:
@MPucheta please resolve the conflicts

nahudalla (approved):

PR #105, opened 2020-02-19
MP/  Fix top bar scrolling with DetailedDiagnose

MPucheta:
preview


nahudalla (changes requested):
@MPucheta This task was about the **Top Bar**, not the **Status Bar**. The Top Bar is _our_ bar at the top of the screen. It should _always be fixed_ and not scroll with the rest of the screen.

nahudalla (approved):

PR #106, opened 2020-02-19
MP/ Remove conditional in amountOfAnswer script

MPucheta:
- Also renamed script to show the new functionality.

@nahudalla, @jmbidev, @ADMontilla  what do you think? should this also change the `updatedAt` field? It's easy to do but I'm having trouble deciding if it's correct to do so. I believe the field should only change due to App related mechanism and not due to script, as scripts may be executed any time and show misguided information.


jmbidev:
I think the `updatedAt` field must be present on all diagnoses even I think we should set it to be equal to createdAt for those that doesnt have it. I think we can use that field after if we want to add local cache. 

Btw removing condition of answered seems fine.

nahudalla (approved):
@MPucheta I agree with what you said about the `updatedAt` being for changes from the App **or the Admin interface**. I also agree with what @jmbidev said.

jmbidev (approved):

PR #108, opened 2020-02-21
JB/ Script to remove images not being referenced by database.

jmbidev:


scripts/removeImagesNotLinked.js
```diff
@@ -0,0 +1,52 @@
...
+  const imageReferences = new Set()
+  const snapshot = await admin.firestore().collectionGroup('requests').get()
+
+  snapshot.docs.forEach(
+    (doc) => {
+      doc.data().imageReferences.forEach(image => {
+        imageReferences.add(image)
+      })
+    }
+  )
```
MPucheta, line 28: Please add a blank before `return`

scripts/removeImagesNotLinked.js
```diff
@@ -0,0 +1,52 @@
...
+  return imageReferences
+}
+
+const getAllStoredImages = async () => {
+  const storedImagesSnapshot = await bucket.getFiles()
+  const storedImagesFiles = storedImagesSnapshot[0]
+
+  return storedImagesFiles.map(image => image.metadata.name.split('/')[1])
```
MPucheta, line 36: I would extract `image.metadata.name.split('/')[1]` into a function that tells what the split is aiming for

Something like `getImageName`, `extractImageName`, `getImageUID` or the like

MPucheta (approved):
Excellent job! Code pretty much reads like text.

I left some non-blocking comments

nahudalla (approved):

PR #110, opened 2020-02-27
AM/keep-login

acures:
Persist Login
- I used `onAuthStateChanged` from firebase for persist login


MPucheta (approved):

admin-src/index.js
```diff
@@ -21,13 +21,19 @@ const App = () => {
 
   const onUserLogin = () => setUserLoggedIn(true)
 
+  useEffect(() => {
+    firebase.auth().onAuthStateChanged((user) => {
+      if (user) {
+        onUserLogin()
+      }
+    })
+  })
```
nahudalla: This will execute **every time** the `App` component renders. Please add an empty dependencies array so that it's only executed when the component is mounted:
```suggestion
  useEffect(() => {
    firebase.auth().onAuthStateChanged((user) => {
      if (user) {
        onUserLogin()
      }
    })
  }, [])
```

nahudalla (changes requested):

nahudalla (approved):

nahudalla:
@ADMontilla please resolve the conflicts.

PR #111, opened 2020-02-28
MP/ Add Password Change screen

MPucheta:
- This includes error handling, some tests, and new generic components that arise from repeated use cases

-  Although new components were created, previous implementations were not changed (as it's not this PR responsibility)

image

image

image

image


nahudalla (approved):
:clap: :muscle:

PR #113, opened 2020-03-03
AM / Set isAdminLastComment

acures:
- Script for set `isLastCommentAdmin` in older requests


scripts/setIsLastCommentAdmin.js
```diff
@@ -0,0 +1,35 @@
...
+          .collection(`users/${doc.ref.parent.parent.id}/requests/${doc.id}/responses`)
+          .orderBy('createdAt', 'desc')
+          .limit(1)
+          .get()
+
+        responses.docs.map((response) => {
+          doc.ref.update({
+            isLastCommentAdmin: response.data().hasOwnProperty('answeredBy')
+          })
+        })
```
nahudalla: Don't use `map` if you're only iterating over an Array. It's meant to **create a new array with a mapping of the values of the previous one**. Use `forEach` if you only need to iterate.

**However**, is it absolutely necessary to iterate here? The query will return only one document because of the `limit(1)`.

**Also**, you should `await` the `update` operation.

scripts/setIsLastCommentAdmin.js
```diff
@@ -0,0 +1,35 @@
+const admin = require('firebase-admin')
+const { initFirebaseAdmin } = require('./utils/initFirebaseAdmin.js')
+
+initFirebaseAdmin()
+
+const setIsLastCommentAdmin = async () => {
```
nahudalla, line 6: What do you think about adding a little bit of logging? This is a script and it'd be useful for the user of the script to know what's happening. Things like `Loading requests...`, `Setting "isLastCommentAdmin: (true/false)" to request id (id)...`, `Finished updating database`.

scripts/setIsLastCommentAdmin.js
```diff
@@ -0,0 +1,35 @@
...
+        })
+      })
+    )
+  } catch (error) {
+    console.log(error.message)
+  }
+}
+
+setIsLastCommentAdmin()
+  .then()
```
nahudalla: Very useful this empty `then` call :smile:

nahudalla (changes requested):

nahudalla (approved):

PR #114, opened 2020-03-09
MP/ Add script and cloud function for number of new request by the week.

MPucheta:
This basically outputs

image

According to this [documentation](https://firebase.google.com/docs/firestore/manage-data/transactions), `FieldValue.increment()` is an atomic operation


functions/onNewRequestUpdateMetrics/onNewRequestUpdateMetrics.function.js
```diff
@@ -0,0 +1,23 @@
...
+    const endDate = currentMoment.endOf('week').format('YYYY-MM-DD')
+
+    admin
+      .firestore()
+      .doc(`metric-amount-of-requests/${startDate}_${endDate}`)
+      .set({
+        startDate,
+        endDate,
+        amountOfRequests: admin.firestore.FieldValue.increment(1)
+      }, { merge: true })
```
nahudalla, line 20: Does this code create a new document initialized with `amountOfRequests: 1` if the referenced document doesn't exist?

scripts/buildWeeklyRequestCountDocuments/index.js
```diff
@@ -0,0 +1,53 @@
...
+      .firestore()
+      .collectionGroup('requests')
+
+    const snapshots = await allRequests.get()
+
+    await Promise.all(
+      snapshots.docs.map(async (doc) => {
+        console.log(`Adding request: ${doc.id} to the count`)
+
+        const data = doc.data()
```
nahudalla: Does this need an `await`? If it doesn't, remove the `async` from this function and the `Promise.all` surrounding it, since you're not doing any async work.

scripts/buildWeeklyRequestCountDocuments/index.js
```diff
@@ -0,0 +1,53 @@
...
+            endDate,
+            amountOfRequests: 1
+          }
+        }
+      })
+    )
+
+    Object.entries(weeklyRequestCreation).forEach(async ([key, data]) => {
+      await admin
+        .firestore()
```
nahudalla: You're not `await`ing the operation at the main level. Change `forEach` with `map` and surround it with a `await Promise.all(...)`

scripts/buildWeeklyRequestCountDocuments/index.js
```diff
@@ -0,0 +1,53 @@
...
+        .doc(`metric-amount-of-requests/${data.startDate}_${data.endDate}`)
+        .set(data, { merge: true })
+    })
+  } catch (error) {
+    console.error(error.message)
+  }
+}
+
+buildRequestCountDocuments()
+  .then()
```
nahudalla: Empty `then` here?

nahudalla (changes requested):

functions/onNewRequestUpdateMetrics/onNewRequestUpdateMetrics.function.js
MPucheta: Yes. `set` creates the document if it doesn't exist and `increment` creates the field with a value of `1` if the field doesn't exists.

https://firebase.google.com/docs/reference/js/firebase.firestore.FieldValue#static-increment

scripts/buildWeeklyRequestCountDocuments/index.js
MPucheta: Yes, logging something here like 'Script ended' executes out of order and it would be weird to only have a `catch` block, as the general contract is to have a resolve/reject handlers even if they are empty.

nahudalla: :thinking: how are you logging here? `then` receives a function. It can't be executed before the promise resolves:
```suggestion
  .then(() => console.log('Operation finished'))
```

nahudalla (approved):

PR #122, opened 2020-04-20
Hp/Add-username-recommendations

hernan-dp:
Screenshot_20200420-181400_Dr Cannabis
Screenshot_20200420-181539_Dr Cannabis
Screenshot_20200420-181713_Dr Cannabis
Screenshot_20200420-181715_Dr Cannabis

-Created Database function in databaseService that receives a array of possible usernames and returns a array with usernames that are already occupied.
-Added a SuggestionBox component in src/components 
-Re factorized Submit Button
-Created a Suggestion component to  chose possible username for the user


src/components/SuggestionBox/index.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import { StyleSheet, View } from 'react-native'
+import Body from '../texts/Body'
+import { theme } from '~/constants'
+import { verticalScale, scale } from 'react-native-size-matters'
```
MPucheta: Use `react-native-size-matters/extend`

src/components/SuggestionBox/index.js
```diff
@@ -0,0 +1,61 @@
...
+    <View style={[suggestionsStyle, style]} {...restProps}>
+      <Body
+        style={[textStyle, style]} {...restProps}
+      >
+        {children}
+      </Body>
+      <SvgXml xml={check} style={styles.icon} />
+    </View>
+  )
+}
```
MPucheta: It's weird to have the same style prop for different components. You then re-apply these styles in the `return` clause.

Separate the style into two: `boxStyle`, `textStyle`.

Then in the `return` use these arrays `style={suggestionStyle}` and `style={textStyle}` (they already got the style prop applied on definition `[styles.suggestionBox, boxStyle]`)

src/components/SuggestionBox/index.js
```diff
@@ -0,0 +1,61 @@
...
+    backgroundColor: theme.colors.white,
+    borderRadius: scale(5),
+    borderColor: '#DADADA',
+    borderBottomWidth: scale(1)
+  },
+  text: {
+    flex: 1,
+    fontFamily: 'Red Hat Display',
+    fontSize: scale(14),
+    lineHeight: verticalScale(15),
```
MPucheta: Remove text styles. Body component has defaults to maintain consistency with the styleguide

image

src/constants/theme.js
```diff
@@ -68,6 +68,10 @@ const fonts = {
   error: {
     fontFamily: 'red-hat-display-medium',
     fontSize: sizes.error
+  },
+  redError: {
+    fontFamily: 'open-sans',
+    fontSize: sizes.error
```
MPucheta: There is no indication that this constant here defines a red color other than it's name, it's also hard to change, for example let's say that a designer later wants all errors to be dark red or brown, this change would cause the `redError` to be another color, it would be confusing to use and will cause changes in any place it is used. Basically prefer more abstract names, which are not described by it's properties but by it's purpose.

Also `open-sans` doesn't exist in Android, use the `error` which has `red-hat-display`.

src/databaseService/index.js
```diff
@@ -50,7 +50,6 @@ const add = async (collectionPath, data) => {
 
 export const usernameAlreadyInUse = async (username, collectionPath = 'users') => {
   const email = await queryEmailFromUsername(username, collectionPath)
-
```
MPucheta: Please rollback this. We always introduce blank lines before the return clause unless it's the only instruction.

src/databaseService/index.js
```diff
@@ -78,6 +77,18 @@ export const queryUsernameFromEmail = async (email, collectionPath = 'users') =>
   }
 }
 
+export const findUsersWithUsernames = async (usernames, collectionPath = 'users') => {
+  try {
+    const querySnapshot = await firebase.firestore().collection(collectionPath).where('username', 'in', usernames).get()
+    if (hasDocuments(querySnapshot)) {
+      const usedUsernameArray = querySnapshot.docs.map(d => d.data().username)
```
MPucheta, line 87: Add a blank line before the `return`

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -33,14 +34,15 @@ const Error = ({ error }) => (
...
       </GrayButton>
     )
   } else {
     return (
       <PrimaryButton
         onPress={onPress}
+        style={styles.submitButton}
       >
         <Description white>Continuar</Description>
       </PrimaryButton>
```
MPucheta: This is old code from before we added the `disabled` prop into `PrimaryButton`.

Please change it to use the prop. It would be similiar to this

```
const SubmitButton = ({ disabled, onPress }) => {
 return (
  <PrimaryButton
   onPress={onPress}
   style={styles.submitButton}
   disabled={disabled}
  >
   <Description white>Continuar</Description>
  </PrimaryButton>
 )
}
```

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -49,40 +51,94 @@ const SubmitButton = ({ disabled, onPress }) => {
...
+        </SuggestionBox>
+      </Button>
+      <Button
+        onPress={() => onSuggestionSelected(suggestions[3])}
+      >
+        <SuggestionBox style={styles.lastSuggestionBox}>
+          {suggestions[3]}
+        </SuggestionBox>
+      </Button>
+    </View>
```
MPucheta: I think this is fine and readable, but someone may argue this could be implemented using `map` function from array.

For example

```
suggestions.map((suggestion, index) => {
return (
 <Button onPress={() => onSuggestionSelected(suggestion)}>
  <SuggestionBox>
    {suggestion}
  </SuggestionBox>
 </Button>
 )
})
```

With a check to add the `lastSuggestionBox` style. Suggestions probably won't grow more than four elements, so I think your solution is okay, but in other cases it would be good to have the `map`, as adding more elements would "scale" automatically.

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -49,40 +51,94 @@ const SubmitButton = ({ disabled, onPress }) => {
...
+      'tricoman',
+      'thegreendoctor',
+      'bluesativa',
+      'Sintabaco',
+      'paulaenverde',
+      'joseliyo',
+      'jamaica97',
+      'cannabisIndica',
+      'cogollosano'
+    ]
```
MPucheta: This is constant, you can move it outside the `UsernameRequest` component.

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -100,10 +156,35 @@ const UsernameRequest = (props) => {
...
+        posibleUsernames[i] = (textInputValue + '_' + Math.floor((Math.random() * 100) + 1))
+        i++
+      }
+      const usedUsernames = await DatabaseService.findUsersWithUsernames(posibleUsernames)
+      if (usedUsernames) {
+        posibleUsernames = posibleUsernames.filter(u => !(usedUsernames.includes(u)))
+      }
+      setSuggestions(posibleUsernames)
+    }
+    if (usernameExists) {
```
MPucheta, line 150: Add a blank line to separate the `if`

src/screens/authentication/UsernameRequest/styles.js
```diff
@@ -6,57 +6,60 @@ export const PLACEHOLDER_TEXT_COLOR = theme.colors.gray
...
-    top: verticalScale(175)
-  },
-  iconBackground: {
-    alignSelf: 'center',
-    position: 'absolute',
-    top: verticalScale(ICON_TOP_DISPLACEMENT_IN_DP + 10),
-    width: moderateScale(62),
-    height: moderateScale(42),
-    backgroundColor: 'rgba(94, 206, 132, 0.2)'
+    display: 'flex',
```
MPucheta: `display: flex` is the default, consider removing it.

src/screens/authentication/UsernameRequest/styles.js
```diff
@@ -6,57 +6,60 @@ export const PLACEHOLDER_TEXT_COLOR = theme.colors.gray
...
+    marginTop: verticalScale(30)
+  },
   error: {
-    alignSelf: 'center',
-    marginBottom: theme.sizes.margin
+    marginLeft: scale(24),
+    alignSelf: 'flex-start',
+    marginBottom: theme.sizes.margin,
+    color: theme.colors.error,
+    fontSize: theme.sizes.error
```
MPucheta: Consider using the Error text component instead of these styles (fontSize, color)

src/screens/authentication/UsernameRequest/styles.js
```diff
@@ -6,57 +6,60 @@ export const PLACEHOLDER_TEXT_COLOR = theme.colors.gray
...
+  suggestionContainer: {
+    borderColor: '#DADADA',
+    borderRadius: scale(5),
+    borderWidth: scale(1)
+  },
+  lastSuggestionBox: {
+    borderBottomWidth: scale(0)
+  },
+  suggestionTitle: {
+    ...theme.fonts.h3
```
MPucheta: I don't see this style being applied, consider removing it as it could be unused code.

src/screens/authentication/UsernameRequest/styles.js
```diff
@@ -6,57 +6,60 @@ export const PLACEHOLDER_TEXT_COLOR = theme.colors.gray
...
+    alignSelf: 'flex-start',
+    marginBottom: theme.sizes.margin,
+    color: theme.colors.error,
+    fontSize: theme.sizes.error
   },
   submitIndicator: {
-    alignSelf: 'center',
     marginBottom: theme.sizes.margin
+  },
+  // Sugestiones
```
MPucheta: Remove this comment please

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -49,40 +51,94 @@ const SubmitButton = ({ disabled, onPress }) => {
...
 )
 
 const UsernameRequest = (props) => {
   const [textInputValue, setTextInputValue] = useState('')
   const [error, setError] = useState(null)
   const [submitting, setSubmitting] = useState(false)
+  const [suggestions, setSuggestions] = useState([null])
+  const [usernameExists, setUsernameExists] = useState(null)
 
   AnalyticsService.setCurrentScreenName('Username Request')
```
MPucheta: Move this inside a useEffect with no dependencies.

```
useEffect(() => {
 AnalyticsService.setCurrentScreenName('Username Request')
}, [])
```

MPucheta (changes requested):
Excellent job Hernan!

I added a couple of comments, mostly code conventions and such. I mainly mark this as a `Request changes` due to the `AnalyticsService` suggestion (we should also apply the same treatment in any screen we see that `AnalyticsService.setCurrentScreen` is outside a `useEffect` hook with no dependencies) and some changes regarding texts styles (prefer to use the defined components instead of manually applying those styles)

src/screens/authentication/UsernameRequest/index.js
hernan-dp: do you separate all blocks by convention (if for return else while etc) or only if's and return?

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -19,70 +42,66 @@ const SubmitIndicator = ({ submitting }) => (
...
   const submitHandler = async () => {
     setSubmitting(true)
     setError(null)
     let newUsernameCreated = false
 
     try {
-      if (await DatabaseService.usernameAlreadyInUse(textInputValue)) {
-        setError('Este nickname ya existe')
+      if (await DatabaseService.usernameAlreadyInUse(textInputValue) || usernamesOcuppied.includes(textInputValue)) {
+        setError('El usuario "' + textInputValue + '" no está disponible')
```
nahudalla:
```suggestion
        setError(`El usuario "${textInputValue}" no está disponible`)
```

src/screens/authentication/UsernameRequest/index.js
nahudalla: All blocks.

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -118,10 +169,19 @@ const UsernameRequest = (props) => {
           onChangeText={(text) => handleInputText(text)}
           value={textInputValue}
         />
-        <Error error={error} />
+        {error &&
+          <Error style={styles.error}>  El usuario {'"' + textInputValue + '"'} no está disponible </Error>}
```
nahudalla:
```suggestion
          <Error style={styles.error}>  El usuario "{textInputValue}" no está disponible </Error>}
```

nahudalla (approved):

src/components/SuggestionBox/index.js
```diff
@@ -0,0 +1,58 @@
...
+    <View style={suggestionsStyle} {...restProps}>
+      <Body
+        style={textStyle} {...restProps}
+      >
+        {children}
+      </Body>
+      <SvgXml xml={check} style={styles.icon} />
+    </View>
+  )
+}
```
MPucheta: Add props `textStyle` and `boxStyle` and change this with:

```
const suggestionsStyles = [
    styles.suggestionsbox,
    boxStyle
  ]

  const textStyles = [
    styles.text,
    textStyle
  ]
```

and then

```
<View style={suggestionsStyles} {...restProps}>
      <Body
        style={textStyles} {...restProps}
      >
```

By doing this you increase the flexibility of the component

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -19,70 +42,66 @@ const SubmitIndicator = ({ submitting }) => (
...
-      style={styles.error}
-    >
-      {error}
-    </Body>
-  )
+const SubmitButton = ({ disabled, onPress, error }) => (
+  <PrimaryButton
+    onPress={onPress}
+    style={error ? { marginTop: verticalScale(20) } : { marginTop: verticalScale(40) }}
+    disabled={error ? true : disabled}
```
MPucheta: This is equivalent to `error || disabled`

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -19,70 +42,66 @@ const SubmitIndicator = ({ submitting }) => (
...
-      secondary
-      style={styles.error}
-    >
-      {error}
-    </Body>
-  )
+const SubmitButton = ({ disabled, onPress, error }) => (
+  <PrimaryButton
+    onPress={onPress}
+    style={error ? { marginTop: verticalScale(20) } : { marginTop: verticalScale(40) }}
```
MPucheta: Consider moving all styling into `styles.js` when it's a big component file like this. Something like `submitWithError` and `submitWithoutError` should suffice.

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -19,70 +42,66 @@ const SubmitIndicator = ({ submitting }) => (
...
+    style={styles.icon}
+    xml={usernameIcon}
+  />
+)
+
+const Suggestions = ({ error, suggestions, onSuggestionSelected }) => (
+  error && (
+    <View style={styles.suggestionContainer}>
+      {suggestions.map((suggestion, index) => {
+        if (index < 4) {
```
MPucheta: This `if` is weird, it's purpose is to prevent the map from returning something. If I want to add more elements to `suggestions` I would have to check the code as to why my new element is not rendering. Please remove it and in any case force `suggestions` to always be a 4-length array

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -19,70 +42,66 @@ const SubmitIndicator = ({ submitting }) => (
...
+          return (
+            <Button onPress={() => onSuggestionSelected(suggestion)} key={index}>
+              <SuggestionBox style={(index === 3) ? { borderBottomWidth: scale(0) } : {}}>
```
MPucheta: This `3` is actually the last index, I dont think it's necessary to change this but consider using computed values from the structure instead of hard coded values. For example `suggestions.length - 1`

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -100,10 +119,42 @@ const UsernameRequest = (props) => {
...
+    AnalyticsService.setCurrentScreenName('Username Request')
+  }, [])
+
+  useEffect(() => {
+    const createSuggestions = async () => {
+      let posibleUsernames = []
+      let i = 0
+
+      while (i < 9) {
+        posibleUsernames[i] = (textInputValue + '_' + Math.floor((Math.random() * 100) + 1))
```
MPucheta: `9` and `100` are magic numbers, which means they serve a purpose but I have to read the code to know what they mean by context. Were they handpicked? They represent a threshold?

I understand the code, but it probably would improve it's readabilty to define

`const UPPER_USERNAME_LIMIT = 100` 
`const NUMBER_OF_POSSIBLE_USERNAMES = 9`

or something like that.

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -100,10 +119,42 @@ const UsernameRequest = (props) => {
...
+    setError(false)
+  }
+
+  useEffect(() => {
+    AnalyticsService.setCurrentScreenName('Username Request')
+  }, [])
+
+  useEffect(() => {
+    const createSuggestions = async () => {
+      let posibleUsernames = []
```
MPucheta: typo in variable, it's `possible`.

src/screens/authentication/UsernameRequest/index.js
```diff
@@ -1,15 +1,38 @@
...
-import { PrimaryButton, GrayButton } from '~/components/buttons'
+import { Body, Description, Subtitle, Error } from '~/components/texts'
+import { PrimaryButton, GrayButton, Button } from '~/components/buttons'
 import * as AuthenticationService from '~/authenticationService'
 import * as DatabaseService from '~/databaseService'
 import NavigationService from '~/navigationService'
 import usernameIcon from '~/assets/images/UsernameRequest/usernameIcon.svg'
 import styles, { PLACEHOLDER_TEXT_COLOR, ICON_WIDTH, ICON_HEIGHT } from './styles'
+import SuggestionBox from '~/components/SuggestionBox'
+import { verticalScale, scale } from 'react-native-size-matters'
```
MPucheta: Use `react-native-size-matters/extend`

MPucheta (approved):

src/screens/authentication/UsernameRequest/index.js
hernan-dp: I tried it  that way but it caused a warning because error is a string and the disable prop only takes booleans

src/screens/authentication/UsernameRequest/index.js
hernan-dp: Yeah the in function in firebase queries only let's you pass 10 parametres so the 9 was from there, i'll make the changes

src/screens/authentication/UsernameRequest/index.js
nahudalla: You can convert it to Boolean: `Boolean(error || disabled)`

PR #123, opened 2020-04-20
MP/Add photos in comment feature

MPucheta:
- Abstracted Gallery logic so it's a reusable component instead of a screen

- Added a FullScreen Carousel component, useful to see pictures in comments

- Added the "add photos in comment" feature, which allows a user to send pictures along with their comment

- Images:

**FullScreenCarousel**
image

**Submit process**

image

**Submit with keyboard**
image

**Gallery inside our screen**
image


src/components/Gallery/index.js
```diff
@@ -0,0 +1,188 @@
...
+
+  const onSubmitPress = useCallback(
+    () => {
+      const length = selectedImages.length
+      if (length >= minSelectionCount && length <= maxSelectionCount) {
+        onSubmit(selectedImages)
+      }
+    },
+    [selectedImages, params]
+  )
```
nahudalla: This is missing `minSelectionCount`, `maxSelectionCount` and `onSubmit` from the dependencies Array. These three variables can change of value because they're props.

src/components/Gallery/index.js
```diff
@@ -0,0 +1,188 @@
...
+        }
+      }
+    },
+    [endCursor, hasMoreImages]
+  )
+
+  const canSelectMoreImages = useCallback(
+    () => selectedImages.length < maxSelectionCount,
+    [selectedImages]
+  )
```
nahudalla: Add `maxSelectionCount` to the dependencies Array.

src/components/Gallery/index.js
```diff
@@ -0,0 +1,188 @@
...
+
+      try {
+        const [image] = await ImageService.openCamera()
+        const newImages = [...images]
+
+        newImages.splice(1, 0, image)
+        await CameraRoll.saveToCameraRoll(image)
+        setImages(newImages)
+      } catch (error) {
+      }
```
nahudalla: Shouldn't we inform the user that something failed?

src/components/Gallery/index.js
```diff
@@ -0,0 +1,188 @@
...
+    if (params.addedImages) {
+      const newImages = [...images]
+      newImages.splice(1, 0, ...params.addedImages)
+      setImages(newImages)
+    }
+
+    if (params.selectedImages) {
+      setSelectedImages([...params.selectedImages])
+    }
+  }, [params])
```
nahudalla, line 148: Add `images` to the dependencies Array.

src/components/Gallery/index.js
```diff
@@ -0,0 +1,188 @@
...
+    return (
+      <ListItem
+        item={item}
+        width={size}
+        height={size}
+        selectedIndex={selectedImages.indexOf(item) + 1}
+        onPress={toggleItem}
+      />
+    )
+  })
```
nahudalla: If you don't specify a dependencies Array, this will return a new callback on every render. The dependencies Array needs to be: `[size, toggleSelect, selectedImages]`

src/screens/MainApp/DetailedDiagnose/components/Comment/index.js
```diff
@@ -1,16 +1,37 @@
...
+const Comment = ({ text, by, date, images }) => {
+  const renderImage = useCallback((image, index) => {
+    return (
+      <ImageButton
+        onPress={() => NavigationService.navigate('FullScreenImagesView', { images, startImage: index, title: by })}
+        source={{ uri: image }}
+        style={styles.image}
+      />
+    )
+  }, [])
```
nahudalla: Add `images` and `by` to the dependencies Array.

src/screens/MainApp/DetailedDiagnose/components/Comment/index.js
```diff
@@ -1,16 +1,37 @@
...
         <Subheader style={styles.by}>{by}</Subheader>
         <Date gray2>{date}</Date>
       </View>
       <Body gray>{text}</Body>
+      {images && (
+        <FlatList
+          data={images}
+          horizontal
+          showsHorizontalScrollIndicator={false}
+          renderItem={({ item, index }) => renderImage(item, index)}
```
nahudalla: By using a lambda function here, you're throwing away all the benefits of using `useCallback` to define `renderImage` :smile: Just destructure the parameters in the callback definition above and pass `renderImage` directly here.

src/screens/MainApp/DetailedDiagnose/components/NewCommentImageList/index.js
```diff
@@ -0,0 +1,42 @@
...
+        source={{ uri }}
+        onPress={onPress}
+        style={styles.image}
+      />
+    </View>
+  )
+}
+
+const NewCommentImageList = ({ images, onImagePress, style }) => {
+  if (!images) return (<></>)
```
nahudalla: `return null` if you don't want to render anything.

src/screens/MainApp/DetailedDiagnose/index.js
```diff
@@ -192,15 +192,32 @@ const DetailedDiagnose = ({ navigation }) => {
...
+      }
+
+      const response = await DatabaseService.addDiagnoseResponse(diagnose.id, responseBody)
       const renderedResponse = await renderResponse(response)
 
       setResponses([renderedResponse, ...responses])
       setAmountOfResponses(v => v + 1)
     } catch (error) {
       setError(error.message)
+      console.log(error)
```
nahudalla: Remove this `console.log`.

src/screens/MainApp/DetailedDiagnose/renderUtilities/index.js
```diff
@@ -12,11 +13,23 @@ const buildDataFromRawResponse = async (response) => {
...
+  const images = []
+
+  if (response.imageReferences) {
+    await Promise.all(
+      response.imageReferences.map(async (imageRef) => {
+        const url = await getURL(imageRef)
+        images.push(url)
+      })
+    )
+  }
```
nahudalla:
```suggestion
  let images = []

  if (response.imageReferences) {
    images = await Promise.all(
      response.imageReferences.map(
        (imageRef) => getURL(imageRef)
      )
    )
  }
```

src/screens/MainApp/MyDiagnoses/screens/FullScreenImagesView/index.js
```diff
@@ -0,0 +1,42 @@
...
+  const [startImage] = useState(navigation.state.params.startImage)
+  const [title] = useState(navigation.state.params.title || '')
+
+  useEffect(() => {
+    AnalyticsService.setCurrentScreenName('Full Screen Images View')
+  }, [])
+
+  const goBack = useCallback(() => {
+    navigation.goBack()
+  })
```
nahudalla:
```suggestion
  const goBack = useCallback(() => {
    navigation.goBack()
  }, [navigation])
```

src/screens/MainApp/MyDiagnoses/screens/FullScreenImagesView/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const renderStatusBar = useCallback(() => {
+    TranslucentStatusBar.setAsCurrent()
+    StatusBar.setTranslucent(true, true)
+  })
```
nahudalla:
```suggestion
  const renderStatusBar = useCallback(() => {
    TranslucentStatusBar.setAsCurrent()
    StatusBar.setTranslucent(true, true)
  }, [])
```

nahudalla (changes requested):
### Great work!

src/components/Gallery/index.js
MPucheta: This causes an infinite loop when `params.addedImage` is truthy, as `setImages` causes the hook to re-run

nahudalla: @MPucheta then don't use the variable `images` because it can be outdated if it's not in the dependencies Array. To set a state based on a calculation from the previous value, you can pass a function:
```javascript
setImages(images => {
  const newImages = [...images]
  
  newImages.splice(1, 0, ...params.addedImages)

  return newImages
})
```

nahudalla (approved):

PR #127, opened 2020-04-28
MP/Add photos in solve request

MPucheta:
This includes some abstractions that were deemed necessary, for example:

- StorageService has a new function to upload images in a batch. It also includes an `onProgress` parameter to fill up a progress bar.

- I encapsulated the image with the X as a standalone component (not the whole list though, that seemed like a wrong abstraction)

image


nahudalla (approved):
@MPucheta please resolve the conflicts.

PR #129, opened 2020-05-04
MP/ Fix gallery flickering while fast scrolling

MPucheta:
- Flickering should be fixed, although documentation in RN doesn't state why it happens. My guess is some form of render problem in unseen `CellRenderer`.

- Added an optimization lambda, the "why" is provided in a comment.


src/components/Gallery/components/ListItem/index.js
```diff
@@ -17,6 +17,10 @@ const ListItem = ({ item, width, height, selectedIndex, onPress }) => {
   const [baseStyle] = useState(StyleSheet.create({ width, height }))
   const [mutableStyle, setMutableStyle] = useState({})
 
+  const handlePress = useCallback(() => {
+    onPress(item)
+  }, [item])
```
nahudalla: I think `onPress` should be in the dependencies array too.
```suggestion
  }, [item, onPress])
```

nahudalla (approved):
@MPucheta please let me know when you resolve the comment below so that I can merge this PR.

abelosorio (approved):

PR #131, opened 2020-05-11
Split Tabs for open diagnoses and resolved diagnoses implemented

hernan-dp:
merge_from_ofoct

- Added fetchSolvedDiagnosesFromCurrentuser  and fetchUnsolvedDiagnosesFromCurrentuser functions to database.service
- Changed NoDiagnoses component to NoDiagnosesScreen
-  Added NoDiagnoses component to show diferent screen depending on the tab
- Removed Scrolling Header


src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -159,33 +213,46 @@ const MyDiagnoses = ({ navigation }) => {
...
-          onRowOpen={handleListItemSwipe}
-          rightOpenValue={scale(-375)}
-        />}
+      <SwipeListView
+        ref={flatListRef}
+        data={diagnoses}
+        extraData={tab}
+        showsVerticalScrollIndicator={false}
+        style={styles.container}
+        ListHeaderComponent={
```
hernan-dp: I tried to put all this in a arrow function that returned the static header and the tab-view all together but it changed the tab speed when changing tabs @nahudalla  @MPucheta

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -159,33 +213,46 @@ const MyDiagnoses = ({ navigation }) => {
...
+            <StaticHeader />
+            <TabView
+              navigationState={{ index, routes }}
+              renderScene={() => null}
+              renderTabBar={renderTabBar}
+              onIndexChange={setIndex}
+              swipeEnabled={false}
+              lazy
+            />
+            {isEmpty(closedDiagnoses) && tab === 'ResolvedDiagnoses' && <NoDiagnoses tab={tab} />}
```
hernan-dp: Couldn't find a better way  to implement this logic because of how swipelistview header contains it, id be thankful if you had any suggestions

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -72,20 +73,62 @@ const MyDiagnoses = ({ navigation }) => {
...
   const diagnoseToRemoveRef = useRef(null)
+  const [index, setIndex] = useState(0)
+  const [tab, setTab] = useState()
+  const [openDiagnoses, setOpenDiagnoses] = useState([])
+  const [closedDiagnoses, setClosedDiagnoses] = useState([])
+
+  const [routes] = useState([
+    { key: 'OpenDiagnoses', title: 'Abiertas' },
+    { key: 'ResolvedDiagnoses', title: 'Resueltas' }
+  ])
```
nahudalla: If you're not modifying the state, it's better to move this to a constant outside of your component.
```javascript
const routes = [
  { key: 'OpenDiagnoses', title: 'Abiertas' },
  { key: 'ResolvedDiagnoses', title: 'Resueltas' }
]

const MyDiagnoses = ({ navigation }) => {
  // here you can use the `routes` constant
}
```

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -72,20 +73,62 @@ const MyDiagnoses = ({ navigation }) => {
...
+            case 'modified': setClosedDiagnoses(v => processModifiedDiagnose(v, doc)); break
+            case 'removed': setClosedDiagnoses(v => processRemovedDiagnose(v, doc)); break
+          }
+        })
+        setDownloadingDiagnoses(false)
+      })
+      return () => { unsuscribe() }
+    },
+    []
+  )
```
nahudalla: Never duplicate code! You're breaking the DRY rule: Don't Repeat Yourself. This `useEffect` can easily be extracted into a custom, generic hook:
```javascript
function useQueryListener ({ fetch, defaultLoadingStatus = false, defaultResults = [] }) {
  const [loading, setLoading] = useState(defaultLoadingStatus)
  const [results, setResults] = useState(defaultResults)

  useEffect(() => {
    const unsuscribe = fetch(async (snapshot) => {
      setLoading(true)
      snapshot.docChanges().map(docChange => {
        const doc = docChange.doc
        const changeType = docChange.type

        switch (changeType) {
          case 'added': setResults(v => processAddedDiagnose(v, doc)); break
          case 'modified': setResults(v => processModifiedDiagnose(v, doc)); break
          case 'removed': setResults(v => processRemovedDiagnose(v, doc)); break
        }
      })
      setLoading(false)
    })

    return () => { unsuscribe() }
  }, [])

  return { loading, results }
}
```

Then, you can use it like this:
```javascript
const MyDiagnoses = ({ navigation }) => {
  const { results: openDiagnoses, loading: openDiagnosesLoading } = useQueryListener({
    fetch: DatabaseService.fetchUnsolvedDiagnosesFromCurrentUser
  })
  const { results: closedDiagnoses, loading: closedDiagnosesLoading } = useQueryListener({
    fetch: DatabaseService.fetchSolvedDiagnosesFromCurrentUser
  })
  const downloadingDiagnoses = openDiagnosesLoading || closedDiagnosesLoading
}
```

src/screens/MainApp/MyDiagnoses/components/NoDiagnoses/index.js
```diff
@@ -1,36 +1,40 @@
 import React from 'react'
 import { SvgXml } from 'react-native-svg'
 import Background from '~/components/Background'
-import { Subtitle, Description } from '~/components/texts'
+import { Description } from '~/components/texts'
 import { PrimaryButton } from '~/components/buttons'
 import NavigationService from '~/navigationService'
-import icon from '~/assets/images/MyDiagnoses/NoDiagnosesIcon.svg'
+import outline from '~/assets/images/DetailedDiagnose/noCommentsYetOutline.svg'
```
MPucheta: Please consider duplicating the icon in the folder `MyDiagnoses/`, so we have the same icon file in two different paths. Changing the name of that icon or replacing it for `DetailedDiagnose` shouldn't break this screen.

src/screens/MainApp/MyDiagnoses/screens/NoDiagnoseScreen/index.js
```diff
@@ -0,0 +1,37 @@
...
+import { PrimaryButton } from '~/components/buttons'
+import NavigationService from '~/navigationService'
+import icon from '~/assets/images/MyDiagnoses/NoDiagnosesIcon.svg'
+import styles, { ICON_WIDTH, ICON_HEIGHT } from './styles'
+
+const goToRequest = () => {
+  NavigationService.navigate('DiagnoseRequest')
+}
+
+const NoDiagnoseScreen = () => (
```
MPucheta: This is not necessary to change, but consider that naming is provided by context too! 

We know a file describes something of a screen because it's inside the `screens/` path. The same is for components, we do not name them `ButtonComponent`, because we know they are in the `components/` folder. `NoDiagnoses` was a good name, `NoDiagnosesYet` is a good one too.

src/screens/MainApp/MyDiagnoses/index.js
MPucheta: Mmm, the best way is what you described, but also moving this in a `useMemo` hook, to provide a `memoized` value that changes according to dependencies (otherwise each render could be creating those components).

Something like this 

image

If the code you submitted works okay/fast do not worry then.

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -159,33 +213,46 @@ const MyDiagnoses = ({ navigation }) => {
...
+              renderTabBar={renderTabBar}
+              onIndexChange={setIndex}
+              swipeEnabled={false}
+              lazy
+            />
+            {isEmpty(closedDiagnoses) && !downloadingDiagnoses && tab === 'ResolvedDiagnoses' && <NoDiagnoses tab={tab} />}
+            {isEmpty(openDiagnoses) && !downloadingDiagnoses && tab === 'OpenDiagnoses' && <NoDiagnoses tab={tab} />}
+          </View>
+        }
+        renderItem={({ item }) => <RenderDiagnose animationDuration={DIAGNOSE_ANIMATION_DURATION} item={item} />}
```
MPucheta: Please check if you can extract this `renderItem` function outside, otherwise each rerender it's re-created.

For example `renderItem = {renderItem}`

and 

```
const renderItem = useCallback(()=>{
    // return RenderDiagnose
},[])
```

src/screens/MainApp/MyDiagnoses/index.js
nahudalla: > I tried to put all this in a arrow function that returned the static header and the tab-view all together but it changed the tab speed when changing tabs @nahudalla @MPucheta

Probably you should have used a memoized arrow function, with `useMemo` or `useCallback`. Anyway, if this works, it's not that bad to pass rendered components as a prop.

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -159,33 +213,46 @@ const MyDiagnoses = ({ navigation }) => {
...
+            <TabView
+              navigationState={{ index, routes }}
+              renderScene={() => null}
+              renderTabBar={renderTabBar}
+              onIndexChange={setIndex}
+              swipeEnabled={false}
+              lazy
+            />
+            {isEmpty(closedDiagnoses) && !downloadingDiagnoses && tab === 'ResolvedDiagnoses' && <NoDiagnoses tab={tab} />}
+            {isEmpty(openDiagnoses) && !downloadingDiagnoses && tab === 'OpenDiagnoses' && <NoDiagnoses tab={tab} />}
```
nahudalla: I think it'd be better to separate the `NoDiagnoses` component into two separate components: `NoResolvedDiagnoses` and `NoOpenDiagnoses`.

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -159,33 +213,46 @@ const MyDiagnoses = ({ navigation }) => {
...
+        onScroll={({ nativeEvent }) => { setScrolling(isScrollWithinSnapThreshold(nativeEvent)) }}
+        renderHiddenItem={(data, rowMap) => (
+          <View style={[styles.hiddenField]}>
+            <SvgButton
+              height={verticalScale(30)}
+              width={scale(21)}
+              svg={trashIcon}
+            />
+          </View>
+        )}
```
MPucheta: You can probably extract this too, the same logic behind `renderItem`.

`renderHiddenItem={renderHiddenItem}`

```
const renderHiddenItem = useCallback((data, rowMap)=>(
          <View style={[styles.hiddenField]}>
            <SvgButton
              height={verticalScale(30)}
              width={scale(21)}
              svg={trashIcon}
            />
          </View>
        ),[])

```

If it doesn't take too much time, please consider changing it (do not spend more than 1 hour on this, we can always come back to iterate it later!)

src/screens/MainApp/MyDiagnoses/components/NoDiagnoses/index.js
```diff
@@ -1,36 +1,40 @@
...
 import styles, { ICON_WIDTH, ICON_HEIGHT } from './styles'
 
 const goToRequest = () => {
   NavigationService.navigate('DiagnoseRequest')
 }
 
-const NoDiagnoses = () => (
+const NoDiagnoses = ({ tab }) => (
```
nahudalla: I already said this in another comment, but I think it'd be better to separate this component into two separate ones: `NoOpenDiagnoses` and `NoResolvedDiagnoses`.

nahudalla (changes requested):

MPucheta (approved):
Good job Hernan!

I left some comments, many are of some possible optimizations. If this works fast enough, do not worry about them, it's just to show you that some re-renders can be avoided with some additional hook calls (many of which also improve readability).

The only thing I would want modified is to have the duplicated icon, so that changing the `DetailedDiagnose` one doesn't propagate to this screen.

src/screens/MainApp/MyDiagnoses/components/NoDiagnoses/index.js
nahudalla: I wouldn't duplicate the icon, it makes the size of the app increase. It should instead be moved to a common folder for both screens, much like the components and hooks: `~/images`

src/screens/MainApp/MyDiagnoses/index.js
```diff
@@ -95,50 +62,16 @@ const MyDiagnoses = ({ navigation }) => {
...
+  const { results: openDiagnoses, loading: openDiagnosesLoading } = useQueryListener({
+    fetch: DatabaseService.fetchUnsolvedDiagnosesFromCurrentUser
+  })
+  const { results: closedDiagnoses, loading: closedDiagnosesLoading } = useQueryListener({
+    fetch: DatabaseService.fetchSolvedDiagnosesFromCurrentUser
+  })
+  const downloadingDiagnoses = openDiagnosesLoading || closedDiagnosesLoading
 
   useEffect(() => {
+    AnalyticsService.setCurrentScreenName('My Diagnoses')
```
nahudalla: This line cannot be in this `useEffect`. It'll be executed whenever any of the dependencies for this hook changes. Move it to its own effect, like this:
```javascript
useEffect(() => {
  AnalyticsService.setCurrentScreenName('My Diagnoses')
}, [])
```

nahudalla (changes requested):

PR #133, opened 2020-05-12
MP/ add screen when no connection is available

MPucheta:
This screen is shown when trying to do an operation that require internet connection:
image

Add decorator to check and navigate to NoConnection screen for internet-dependant operations

A check for connection is done, if there is not internet, we navigate to NoConnection screen (which needs to be in the routing definition). If there is internet, then the normal function can be executed the same way it was executed before.

This is somewhat of an anti-pattern as the decorators use navigator/routing logic, but it was the best way to provide a DRY solution while maintaining a good readability, if you want this feature, you just decorate the corresponding function.

Alternative was to pollute the code with the network listener hook and manually navigate to the screen, which gives more insight on whats happening, but it clutters the code a lot.


hernan-dp (approved):
Great Work! I learned a lot about connectivity from your code

